PR #84, opened 2015-01-25
Django 1.6 Upgrade and Python requirement upgrade

JoeJasinski:
This pull request is to upgrade to Django 1.6 in addition to upgrading a variety of Python requirements.  (I held off on a 1.7 update since a 1.7 update might require crafting new django-style migrations and testing with actual data.) 

Note: I pulled in changes from the master branch, so those are included in the diff as well. 

## Install Updates 

    pip install -r requirements.txt
    ./manage.py migrate # needed for flatblocks
    
You need to generate new recaptcha keys in the new recaptcha format.  Then export the following variables.

    export NORECAPTCHA_SITE_KEY=your_recaptcha_public_key
    export NORECAPTCHA_SECRET_KEY=your_recaptcha_private_key 

## Summary ##
- updated Django to 1.6
    - changed url tags to new syntax
    - changed urls import to the new location
    - use the TemplateView in place of direct_to_template
    - moved the manage.py up one level to support 
    - renamed the import paths for the 1st party packages to support the manage.py move
    - added ALLOWED_HOSTS to settings
- updated most of the python requirements to their newest version
- replaced django-gravatar with django-gravatar2, which supports newer django versions
- replaced django-recaptcha with django-nocaptcha-recaptcha which supports google's new recaptcha api
 - added some more environment variables to settings.py to make it easier to testing email easier


chipy_org/templates/boxes/box_create.html
```diff
@@ -1,5 +1,5 @@
 <div class="content-box">
-    <form id="edit_form" method="POST" class="in_facebox" action="{% url box_create label %}">
+    <form id="edit_form" method="POST" class="in_facebox" action="{% url 'box_create label' %}">
```
emperorcezar: Why not just wrap `box_create` in quotes?

emperorcezar: I'm also not 100% sure what the box template are for.

chipy_org/apps/meetings/serializers.py
```diff
@@ -4,7 +4,7 @@
 
 
 class PresentorSerializer(serializers.ModelSerializer):
-    email = serializers.SerializerMethodField('get_email')
```
emperorcezar: My understanding of Rest Framework is that this won't work after removing what method to call?

chipy_org/deploy/wsgi.py
```diff
@@ -1,4 +1,14 @@
...
+"""
+WSGI config for chipy2 project.
+
+It exposes the WSGI callable as a module-level variable named ``application``.
+
+For more information on this file, see
+https://docs.djangoproject.com/en/1.6/howto/deployment/wsgi/
+"""
 
-application = Cling(get_wsgi_application())
```
emperorcezar: We were using Cling to serve up the static assets. Why remove it?

emperorcezar:
Good work.

I put a few questions in for things that might need fixing. You'll also need to pull master back into your branch so this is merge-able.

chipy_org/deploy/wsgi.py
JoeJasinski: Thanks for the clarification on that.  I'm adding this back.  I must have overwritten this file when I copied in the new Django project structure.

chipy_org/apps/meetings/serializers.py
JoeJasinski: Rest framework was throwing this error when I included that method call.  Removing the call seemed to fix it and I think I see the appropriate email address in the meetings list api view.  Maybe this is a change in rest framework 3?  

    AssertionError at /api/meetings/
    It is redundant to specify `get_email` on SerializerMethodField 'email' in serializer 'PresentorSerializer', because it is the same as the default method name. Remove the `method_name` argument.

JoeJasinski:
Thank you for having a look at this Cezar. I've merged in the recent changes from master and have clicked through the application's various features and it was working as far as I could tell (and I ran the 3 existing unit tests which passed).  I've responded to the comments as well. 

I'm sure how to best test the new meetup sync features, though from the look of it, it doesn't seem like the upgrade would impact those modules.  

I'm happy to make other adjustments as you would like.  Thanks again.

chipy_org/apps/meetings/serializers.py
emperorcezar: Thanks. I haven't touched Rest Framework 3. Good to know.

Weird it throws an `AssertionError` though. Seems like that would be much more suited to a warning.

emperorcezar:
Looks good. I'll deploy when I'm back on solid internet, so probably a few days.
:+1:

JoeJasinski:
Sound good.  Let me know if can be of any assistance with that.

PR #87, opened 2015-02-12
Mismatches quotes breaking rsvp form

selecsosi:
Was trying to sign up tonight and couldn't get the form to submit. Investigated a bit and found this being parsed with an exception on chrome as an illegal token.


emperorcezar:
Thanks for the fix!

emperorcezar:
...and deployed.

PR #101, opened 2015-09-25
Feature/sponsor

JoeJasinski:
I believe that this is ready for merging.  Let me know if you have any feedback. 

### Features

 - updated gitignore
 - README updates
 - New GeneralSponsor list on homepage
 - New Sponsor Detail page
 - New Past Meeting view
 - New Sponsor Admin section
 - New Sponsor Models
 - Replace dj_static with django-whitenoise
 - Changes to ALLOWED_HOSTS
 - S3 and Local hosting media hosting enabled
 - Sorl Thumbnail added for Sponsor image thumbnails
 - New Footer links for Organizers and Code of Conduct
 - New Menu item for Sponsors
 - Convert STATIC_URL template variable to static template tag (per Django's preference)
 
Updated Packages
 - Pillow
 - boto
 - django-flatblocks
 - feedparser 
 - httplib2
 - psycopg2
 - requests 
 - sorl-thumbanil (new)
 - whitenoise (new)

#### Set environment variable on Heroku

    heroku config:set DEBUG=False
    heroku config:set USE_S3="True"
    heroku config:set AWS_ACCESS_KEY_ID=""
    heroku config:set AWS_SECRET_ACCESS_KEY=""
    heroku config:set AWS_STORAGE_BUCKET_NAME=""

#### Collectstatic

    heroku run python manage.py collectstatic --noinput

#### Set sync and migrate the database

    heroku run python manage.py syncdb
    heroku run python manage.py migrate


chipy_org/settings.py
```diff
@@ -88,16 +87,34 @@ def env_list(key, defaults=[], delimiter=','):
...
+if USE_S3:
+    AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID')
+    AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY')
+    AWS_STORAGE_BUCKET_NAME = os.environ.get('AWS_STORAGE_BUCKET_NAME')
+    AWS_QUERYSTRING_AUTH = False
+    AWS_HEADERS = {
+        'Cache-Control': 'max-age=86400',
+    }
+    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
+    #STATICFILES_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
```
emperorcezar: Why the commented out code?

emperorcezar:
Couple questions:

- How are the sponsor logos pushed to S3?
- How is the `url` on the `Sponsor` model populated?
- Is this dependent on `collect_static` _after_ the heroku push? I ask because the filesystem on Heroku is ephemeral and anything collected will disappear on dyno restart.

Otherwise the code looks good and a big thanks for doing this work!

JoeJasinski:
Thank you for your feedback.

  - The sponsor logos are uploaded to S3 using django-storages.  This manages the upload transparently, as it takes advantage of django's storage backend system. 
  - All of the fields on the Sponsor model are managed by the Django admin.  The url field just adds a link to the sponsor's website.  The logo is uploaded via the admin as well and handled by django-storages.
  - Currently, collectstatic is only used for static media that's contained within the project itself (css/js/images in the project itself).  I believe  that collectstatic is run by heroku upon dyno start.  All of the uploaded media, such as the logos, do not rely on collectstatic.  Those are directly pushed to s3 and are hosted there. Here's Heroku's article about it https://devcenter.heroku.com/articles/django-assets

I'd be glad to discuss this more to make sure all of your questions are answered.

JoeJasinski:
Update: if desired, all of the static files can also be hosted on s3, but that might not be necessary and would add to collectstatic runtimes.

chipy_org/settings.py
JoeJasinski: If we want to host staticfiles on S3, we can uncomment that.  I think we are OK hosting the staticfiles on Heroku since that's all stuff that's generated by the codebase and S3, and my understanding is that Heroku runs a collectstatic on each provision.

emperorcezar:
Durp. Been a long week. I missed the `ImageField`, and you are right about Heroku and collectstatic. Thanks, this is good.

:+1:

emperorcezar:
Will deploy thing in the next couple days when it's not late and my head isn't fried.

JoeJasinski:
Sounds good!

Issue #97, opened 2015-06-14
accuracy on RSVP


brianray:
We should be sure to honor +1 from meetup.com.

emperorcezar:
You'll have to add more details than this.

brianray:
The problem is if we send our rsvp count to a venue, it does not account for +1's

emperorcezar:
Ah, the plus 1's I read this wrong. Thanks!

----------------------------------------------------------------------------------
Father, dabbling at being a Python developer specializing in Django,
Cyclist, and home brewer.

PR #117, opened 2015-12-21
Meetup sync -- add guest count + add user's real name

tanyaschlusser:
To address issue #97 -- adds the guest count to the RSVP list. Also adds the Meetup user's real name if the field is available.


emperorcezar:
:+1:

PR #142, opened 2016-08-17
Fix #141 - Add Link to Membership Page if present

### LINKED ISSUE #141: Link to Mentorship Content
Right now there is no obvious way to connect to the [Mentorship content](http://www.chipy.org/pages/sigs/mentorship/). It would be great to make that a prominent link as it is one of the easier ways to plug in and get more involved in the Python community.

raymondberg:
## Resolves #141 - Add Mentorship Link

There doesn't seem to be an easy way to resolve a link to the flatpages interfacebecause we prefix url matching with `^pages` before routing to the flatpages wildcard matcher and there's no namespacing option that I found useful. 

Additionally, I couldn't find a way to directly resolve the flatpage url based on a single matcher. The flatpages api only exposes [a single custom tag](https://github.com/django/django/blob/c1aec0feda73ede09503192a66f973598aef901d/django/contrib/flatpages/templatetags/flatpages.py) which pulls back an array of results.

 Basically:
1. Look up tags that match the mentorship sig location expected
2. Print links for each (i.e. only the exact one)

This is better, possibly, than a static link because it's not a broken link on local installs.

### DIFF
--- a/chipy_org/templates/site_base.html
+++ b/chipy_org/templates/site_base.html
@@ -1,6 +1,7 @@
 {% extends "theme_base.html" %}
 
 {% load i18n staticfiles %}
+{% load flatpages %}
 
 {% block head_title %}{% block title %}{% endblock %}{% endblock %}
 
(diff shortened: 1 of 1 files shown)


emperorcezar:
Looks good 👍

PR #147, opened 2017-04-18
Add Docker Dev guide

raymondberg:
This is a basic docker guide for folks who want to get a stream-lined startup with docker. Not perfect, but it's a good start.


raymondberg:
Thanks to @adeolapure for helping with the testing.

PR #153, opened 2017-09-07
make the talk submission for a bit better and avoid "emailless" submissions

JoeJasinski:
This pull request enhances the talk submission form:

- [x] FIX: the glaring issue where emails/names are missing from talk submissions.  Now email and name is a required and it pre-populates the form field from the user record if available.  If the user record has an empty email, this will also update that email address. 
- [x] add an optional Notes field for speakers to enter non-public context about their talk
- [x] update the layout a bit, so it makes better use of space on the page
- [x] added helptext to some of the fields for better explanation


JoeJasinski:
Deployed

PR #159, opened 2018-05-15
Replace members header link with code of conduct

raymondberg:
We discussed this, but the Members page isn't useful for most users and the code of conduct is.

It also feels better to put it right after `About` so it's front and center.

@JoeJasinski


raymondberg:
I had brought it up at a board meeting a while back and Joe the group was fine with it. We didn't vote on it.

Issue #162, opened 2018-11-08
Add a way to close RSVPs on a meeting


tathagata:
Without this feature there can be a case when people RSVP on the site after we have sent the list to the host.

JoeJasinski:
PR https://github.com/chicagopython/chipy.org/pull/163

alysivji:
Looks like this is complete. Can we close?

PR #163, opened 2018-11-14
WIP - Feature/rsvp overhall

JoeJasinski:
This is to address several issues. 
 1. We want to be able to export a list of chipy-registered users as a link from the admin interface
 2. We want to be able to disable registrations at a specific date in time
 3. We want the homepage to clearly indicate if/when the registration is closed

Some additional changes that I made
 1. I created csv exports for the user list
 1a. The csv exports are split into public and private lists.  This way we can view the list four our own purposes, but provide an anonymous secret link to our hosts without exposing more sensitive user info. 
 1b. We try to split out the "full name" field into first and last name files for easier sorting in the report.
 2. I re-organized the registration view. It was getting kind of bloated.  It's still bloated, but I'd like to think a bit more organized. 
 3. I fleshed out the docker development environment.  

This addresses:
https://github.com/chicagopython/chipy.org/issues/162

This requires one database migration.


docker/Dockerfile
```diff
@@ -1,19 +1,38 @@
-FROM python:2.7
+FROM python:2.7.15
```
emperorcezar: Any reason why to lock into a minor version?

JoeJasinski: It's unlikely, but something could shift slightly if we don't pin it (for example, python 2.7.9 - ish changed the way that ssl worked a bit). Also it ensures we are using the exact version of python that we are using in Heroku. 

However, it's unlikely to be an issue, and I have no problem removing the minor version pin if others feel strongly about it.

JoeJasinski:
Before I merge and deploy this, I discovered a small bug with the user csv export when running with real data. One of the users has a unicode character in their name which breaks the csv export.  I think I just need to decode it. 

```
Exception Value: 'ascii' codec can't encode character u'\xed' in position 9: ordinal not in range(128)
```

JoeJasinski:
I ended up just encoding the csv row values

JoeJasinski:
This has been deployed

PR #169, opened 2019-02-17
Feature/rsvp sig

JoeJasinski:
This adds the ability to register for SIGs on the SIG detail page. This involved breaking apart the existing RSVP logic since it was only designed to work on the homepage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,18 @@
-Chipy.org
-=========
+# Chipy.org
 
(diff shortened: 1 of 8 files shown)


JoeJasinski:
Thanks for looking at it.  Here's a url to my dev env. (note this also has the changes from my other PR as well)
https://chipy-dev.herokuapp.com/

PR #170, opened 2019-02-17
Feature/add test framework

JoeJasinski:
This depends on https://github.com/chicagopython/chipy.org/pull/169

This is to integrate pytest into chipy.org.  
This also add basic smoke tests for some of our common views - far from comprehensive, but a start.
I've fixed or whitelisted a lot of pylint errors.  I'd like this to be a phase gate when deploying changes.
I've added a CircleCI config to build, lint, and test


JoeJasinski:
@emperorcezar, when you have a chance, mind looking at this?

docker/Dockerfile
```diff
@@ -26,7 +26,8 @@ USER app
 
 # Install python packages
 ADD requirements.txt ${SITE_DIR}requirements.txt
-RUN pip install --user -r ${SITE_DIR}requirements.txt
+ADD requirements_test.txt ${SITE_DIR}requirements_test.txt
```
raymondberg: Would it make sense to switch to another Dockerfile for test vs sharing (e.g. `Dockerfile.test`)? The fear being that if we do use a container-deploy service down the road we'll have to remember to pull out the test stuff from our current Docker image.

JoeJasinski: I went with the single dockerfile for build simplicity. The few testing requirements are in that separate requirements_test.txt file, so they should be pretty easy to identify and pull out down the line if needed. 

Currently, the docker container is used only for local development and figured the testing tools might be beneficial for local dev (vs having to install them each time you start the container, or having to build two docker images, one based on the other).

emperorcezar:
This looks pretty good, sorry I'm just looking at it tonight.

docker/Dockerfile
emperorcezar: I've been thinking about using setup.py and `extras_require` for a bit now. Allows you to do `pip install -e .[test]`

PR #181, opened 2019-03-20
various cleanup items for talk

JoeJasinski:


chipy_org/settings.py
```diff
@@ -9,10 +9,11 @@
...
     """Retrieves env vars and makes Python boolean replacements"""
     val = os.environ.get(key, default)
-    if val in ('True', 'true'):
-        val = True
-    elif val in ('False', 'false'):
-        val = False
+    if isinstance(val, str):
+        if val.lower() in ('t', 'true'):
+            val = True
+        elif val.lower() in ['f', 'false']:
```
tathagata: nit: ['f', 'false'] -> ('f', 'false') to align with line 13. 
super-nit: may be the lower() call can be a separate variable

docker/docker.env.sample
```diff
@@ -44,20 +44,20 @@ EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
 # SMTP email backend and settings
 #EMAIL_BACKEND='django.core.mail.backends.smtp.EmailBackend'
```
tathagata: nit: Looks like this variable is repeated - notice the space in line 43

JoeJasinski: Thanks, I've removed the spaces around the equals sign. The variable is repeated to offer the user a choice of whether to use the smtp email backend or the console backend. I've added some wording in the file to clarify that.

PR #191, opened 2019-03-27
adding some css to make mobile registration modal responsive

obf73:


tathagata:
@obf73 thank you so much for fixing the bug! @JoeJasinski I just rsvped from safari on iphone! This is great!

Issue #183, opened 2019-03-26
Update check list for talk submission to include code of conduct references


tathagata:
As an organizer
in order to prevent code of conduct violation by speakers
I would like the talk submission page to include relevant references to out code of conduct

related: #111

tathagata:
@raymondberg please take a look if you would want any additional details.

raymondberg:
❤️  this

Something like:

> [ ] I have read ChiPy's Code of Conduct and will ensure my presentation and associated materials (slides, handouts, etc...) do not violate the Code of Conduct.

PR #193, opened 2019-03-31
Issue #183 - Add Code of Conduct to talk proposals

raymondberg:
## Problem

We need presenters to know about our Code of Conduct and ensure they know they need to abide by it.

## Solution

Add abbreviated callouts to the legal language about speaker agreement and talk releases.

### Screenshot

Screenshot 2019-03-31 17 58 37


JoeJasinski:
Perhaps we can link this to our existing code of conduct on github? Or use the CMS to add the content. That way we can make editorial changes without having a PR.

raymondberg:
I thought about that, but it felt like nobody would click it. Like a EULA.

JoeJasinski:
If you use the flatblocks application (already installed), you can just add a placeholder for text. It would display the same was as pictured but we could have the text stored in the CMS. We currently use it in several places on the site, include the speaker submission page.

JoeJasinski:
Here's an example:
https://github.com/chicagopython/chipy.org/blob/master/chipy_org/apps/meetings/templates/meetings/propose_topic.html#L17

raymondberg:
Verrrrry nice. Awesome, I'll check it out!

raymondberg:
Nice, merging since comments were incorporated!

Issue #167, opened 2018-12-30
Meeting Description field needs to support rich text formatting


tathagata:
As an event organizer, I want to include rich text in the meeting descriptions, so that I can include links and other formatting such that it makes reading easier.

Project Night is moving to chipy.org for RSVPs. We need to provide multiple instructions for our attendees - on what to bring, what to expect, agenda/timeline, FAQs. example:
https://www.meetup.com/_ChiPy_/events/gdqnkqyzcbwb/
The current description field does not support the formatting needed for organizing such long content and it renders it as a blob of text, very difficult to read.
Supporting rich-text would allow adding the description properly.

JoeJasinski:
Projects to consider:
https://djangopackages.org/packages/p/django-ckeditor/
https://djangopackages.org/packages/p/django-tinymce/

PR #196, opened 2019-05-24
#167 - Meeting Description field needs to support rich text formatting 

smandekar1:
Added rich text functionality.  Allows admin to add rich text to meeting description.  Also homepage display upcoming events(past events) in rich text. 

image
image

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -1,0 +1,111 @@
+import random
+import string
+from django.contrib import admin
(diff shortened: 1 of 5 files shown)


JoeJasinski:
This looks very nice!

raymondberg:
This is the bulk of the work for #167

PR #197, opened 2019-05-27
Document a tagging strategy

raymondberg:
## Problem

We don't currently tag our releases, which makes it difficult to track which code has been deployed to production. 

## Solution

Implement a light-weight tagging strategy to help watchers without prod access see what is out in production.


emperorcezar:
While I don't mind this. I will note that `heroku releases` will give you the sha of the deployment.

PR #200, opened 2019-07-20
Python3 upgrade

raymondberg:
## Problem

We need some upgrades to our stack! Python2 is dying. 

## Solution

Rather than jumping straight to 3.8, I moved us up to 3.4. **Yes, this version is also deprecated**, but should represent one of the larger jumps for us to upgrade to the lastest and greatest across the board. Hopefully, we can do an easy upgrade from 3.4 to at least 3.6 soon after we verify this upgrade works. Django itself is also slated for upgrade, so expect a lot!

The only real issues I ran into were around wsgiref and oauth providers. wsgiref is now baked into python3, so that's easy. Oauth is tougher.

### Oauth

In this PR we are moving [from django-social-auth to its successor python-social-auth](https://python-social-auth.readthedocs.io/en/latest/configuration/porting_from_dsa.html#porting-from-django-social-auth). This isn't the last we need! We also need to move through [from python-social-auth to split auth](https://github.com/omab/python-social-auth/blob/master/MIGRATING_TO_SOCIAL.md), but it seems like that can be a separate upgrade. I'd prefer to get on latest python and then go do a tech refresh on all of our imports.

I've tested and debugged issues to get the same first click behavior as we have in prod, but I need to either deploy this somewhere to test....or risk a prod deploy with a fast follow to fix any lingering issues. Given the amount of time it takes I'm leaning toward that, but I'll talk to anybody who leaves a "LETS CHAT ABOUT DEPLOYS" comment on this PR.  Let's not talk out deploy strategies here.


ZaxR:
LETS CHAT ABOUT DEPLOYS

chipy_org/settings.py
```diff
@@ -170,13 +170,10 @@ def env_list(key, defaults=None, delimiter=','):
 LOGIN_ERROR_URL = '/'
 
 AUTHENTICATION_BACKENDS = [
-    'social_auth.backends.twitter.TwitterBackend',
-    'social_auth.backends.facebook.FacebookBackend',
-    'social_auth.backends.google.GoogleOAuth2Backend',
-    'social_auth.backends.browserid.BrowserIDBackend',
-    'social_auth.backends.contrib.linkedin.LinkedinBackend',
```
ZaxR: Are we dropping facebook and linkedin?

raymondberg: Yeah, we only support openid, google, githhub right now.

raymondberg:
@eevelweezel @adambain-vokal @tathagata @patrickboland4 @alysivji @JoeJasinski

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:2.7.15
+FROM python:3.4.10
```
elmq0022: Covers local docker build

runtime.txt
```diff
@@ -1 +1 @@
-python-2.7.15
+python-3.4.10
```
elmq0022: Covers Heroku deploy

Issue #173, opened 2019-03-15
The upcoming events on the homepage should not be so long


JoeJasinski:
As a ChiPy event scheduler, I want to be able to enter a lot of detailed information for an event, without it consuming so much space on the homepage and instead show it on the event detail page, so that the esthetic of the homepage is not disrupted. 

Currently on the ChiPy homepage "Upcoming Events" section, if an admin enters a really large description, the box stretches really tall and takes up too much space.  Two options to solve:

 1. truncate the paragraph if it's too long. Really easy and effective. See below
    https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#truncatechars
    https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#truncatechars-html
    https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#truncatewords-html
 2. use the same javascript [more] dropdown that we use for "This Month's Topics" on the hompage.

smandekar1:
fixed with truncatewords_html on homepage.  Believe this can be closed

elmq0022:
Awesome 👍

PR #203, opened 2019-10-17
Description safe

smandekar1:
This should handle safe text for descriptions in the: 

1.  Homepage - This month's topic section
2.  Past meetings page
3.  emails that are sent for confirmations, etc.


homepage.html
```diff
@@ -136,7 +136,7 @@ <h3>In the Loop</h3>
                 {% if next_meeting %}
 
                 {% if next_meeting.description %}
-                <p itemprop="description">{{ next_meeting.description|truncatewords:80 }}</p>
+                <p itemprop="description">{{ next_meeting.description|truncatewords:80|safe }}</p>
```
raymondberg: I don't think you want to use safe here. This would render the first few starter HTML tags but then wreck the rest of the page, right?

smandekar1: Seems to render ok, cannot crash it.  Can take safe out but then the p tags show.  seems like main meetings don't have a description anyway though.  

is there an extra homepage.html file in addition to one under templates? when I remove it, I don't see any problems.   

seeing opposite Javascript behavior for the [more]/[less] functionality under This Month's Topic section with |safe, could not figure that out 
Screen Shot 2019-10-24 at 11 52 28 AM

raymondberg:
We talked a bit about this and decided to go with something a little more stripped down. Maybe shortening non-formatted text and linking to the rich HTML on the view page...just like the main meeting description.

chipy_org/apps/sponsors/templates/sponsors/sponsor_list.html
```diff
@@ -5,6 +5,10 @@
 {% block head_title %}{% trans "Sponsor List" %}{% endblock %}
 
 {% block body %}
+
+<h2>{{meeting_attendees}} Python Enthusiasts Attended ChiPy Meetings in the Last Year!!</h2>
```
elmq0022: Hey Sam.  I think this goes with a different pull request you are working on.

homepage.html
```diff
@@ -228,9 +228,9 @@ <h3>This Month's Topics</h3>
                             {{presenter.name}} {% if not forloop.last %}, {% endif %}
                           {% endfor %}<br />
                       {% if topic.experience_level %}
-                      Experience Level: {{ topic.get_experience_level_display }}<br />
+                      BUTTS Experience Level: {{ topic.get_experience_level_display }}<br />
```
elmq0022: Maybe remove BUTTS from here.

elmq0022: There is a JavaScript function in the home page associated with trimming and toggling the description.  If we don't need it anymore could you delete it?

raymondberg: MAAAYBE :D

raymondberg:
I will approve after the small changes Aaron advised!

chipy_org/apps/sponsors/templates/sponsors/sponsor_list.html
smandekar1: Deleting those lines

homepage.html
smandekar1: This is an extra homepage.html so I'm deleting it.  The real one is in chipy_org/templates/homepage.html.  Also removing the unneeded Javascript.

elmq0022:
As a last check I will do some user testing over the next few days.  Will approve if everything checks out. Not expecting anything.  Updates look good.  Thanks!

smandekar1:
Thanks for the notes!

chipy_org/apps/meetings/templates/meetings/emails/meeting_topic_submitted.html
```diff
@@ -5,4 +5,4 @@
 <p>Presenters: {% for presenter in topic.presentors.all %}{{ presenter.name }} ({{ presenter.email }}) - {{ presenter.user.id }} {% endfor %}</p>
 <p>Meeting: {{ topic.meeting }}</p>
 <p>Length: {{ topic.length }}</p>
-<p>Description: {{ topic.description }}</p>
+<p>Description: {{ topic.description|safe }}</p>
```
elmq0022: topic.description is generated from user input and should not be marked "safe".  Please look for a way to sanitize (remove bad html tags such as script) from the user input.

chipy_org/templates/homepage.html
```diff
@@ -230,7 +230,8 @@ <h3>This Month's Topics</h3>
                       {% if topic.experience_level %}
                       Experience Level: {{ topic.get_experience_level_display }}<br />
                       {% endif %}
-                      <span  class="more">{{ topic.description }}</span>
+                      {{ topic.description|striptags|truncatewords:80|safe }}<br />
```
elmq0022: if you can sanitize the html from the user could you use
`{{ topic.description|truncatewords_html:80|safe }}<br />`
instead of 
`{{ topic.description|striptags|truncatewords:80|safe }}<br />`
the output of striptags is not guaranteed to be "safe"

elmq0022:
Sam, you may want to look at the bleach library which whitelists specific tags.

chipy_org/templates/homepage.html
```diff
@@ -230,7 +232,7 @@ <h3>This Month's Topics</h3>
                       {% if topic.experience_level %}
                       Experience Level: {{ topic.get_experience_level_display }}<br />
                       {% endif %}
-                      {{ topic.description|striptags|truncatewords:80|safe }}<br />
+                      {{ topic.description|striptags|truncatewords:80|bleach|safe }}<br />
```
elmq0022: A minor enhancement is to remove the striptags filter and use truncatewords_html.  This leaves any formatting from allowed tags intact.

smandekar1: Nice, I believe we were looking to have the homepage stripped down and the details page to have the rich text.

smandekar1: Thanks Aaron!!😀

elmq0022:
Looks like issue #173 is related.  Maybe you could add the truncatewords-htm tag here as well and close that issue.

PR #204, opened 2019-10-22
Add SponsorGroup and render sponsors according to level

raymondberg:
## Problem
We are trying to expose our sponsorship levels to be a bit more visible
on the site. Right now the sponsors are all listed in alphabetical order
with no precedence for sponsor levels. We want to call out significant
sponsors first (for example if Zoro was our top sponsor), rather than
listing them last because of the alphabetical.

## Solution

This introduces a new model called SponsorGroup to let us be flexible
about naming. We can introduce new groups or change orders with the
list_priority as need be, and the groups will only be displayed if
there's a corresponding Sponsor in the Group.

Example of Before:

Screenshot 2019-10-21 22 04 58

Example of After:

Screenshot 2019-10-21 22 37 35

@JoeJasinski @ZaxR @eevelweezel

### DIFF
--- a/chipy_org/apps/sponsors/models.py
+++ b/chipy_org/apps/sponsors/models.py
@@ -39,6 +39,14 @@ class GeneralSponsor(models.Model):
         ordering = ["sponsor__name"]
 
 
+class SponsorGroup(models.Model):
+    name = models.CharField(max_length=80)
+    list_priority = models.IntegerField(default=5)
+
+    def __str__(self):
+        return self.name
+
+
 class Sponsor(models.Model):
 
     name = models.CharField(max_length=80)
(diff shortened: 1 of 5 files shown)


chipy_org/apps/sponsors/models.py
```diff
@@ -39,6 +39,14 @@ class Meta:
...
 
 
+class SponsorGroup(models.Model):
+    name = models.CharField(max_length=80)
+    list_priority = models.IntegerField(default=5)
+
+    def __str__(self):
+        return self.name
+
+
```
elmq0022: Ray, I wonder if a [choice field](https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices) is more appropriate than a very flexible model.

My concern is someone could either a) author a similar group or b) create a SponsorGroup with the same list_priority.  

The flexibility is nice to have, but maybe unneeded if you already have well defined groups in mind.  

The counter argument is adding a new SponsorGroup would not require a deploy.

**Approved**

elmq0022:
One other question, do we want to add this to chipy.org/sponsors/list as well?

PR #209, opened 2019-11-20
Added make shell Command for Local Development

elmq0022:
I found runing manage.py and pytest commands via docker-compose verbose.  This pull request adds a `shell` command to the Makefile.  The command opens a shell in the web container allowing local use of manage.py, pytest, pylint, etc. directly.


smandekar1:
looks fine to me, awesome!

PR #210, opened 2019-11-21
replaced reverse_lazy with reverse in all tests

elmq0022:
The function reverse_lazy does not work with the planned python3-update branch planned by @raymondberg.  I've updated these tests to use the reverse function which does work with the planned python 3.4 update.


raymondberg:
I just pulled your commit to the upgrade PR https://github.com/chicagopython/chipy.org/pull/200

I'm going to close this down.

Also, you are sitting right next to me :D

PR #217, opened 2019-11-30
python social auth fixes for py3

elmq0022:
@raymondberg  I have updates here for the python3 upgrade related to python social auth  

The pipeline was broken, so I fixed those.  Also the import was incorrect in the chipy library, so I fixed that as well.  

I moved all the settings together so it's not so annoying to look through.  

I tested locally against google oauth2 and was able to login, update my profile, submit a talk, and retrieve my talk.  

The only think I see that is still broken is the login button in the upper right corner does not display the user name when someone is logged in as it does now.

I may try to fix that later, but would appreciate it if you could or other web-guild members could review the changes.


chipy_org/settings.py
```diff
@@ -29,7 +28,7 @@ def env_list(key, defaults=None, delimiter=','):
 
 PROJECT_ROOT = os.path.abspath(os.path.dirname(__file__))
 
-DEBUG = env_var('DEBUG', False)
+DEBUG = env_var('DEBUG', True)
```
raymondberg: I don't think this should stay in.

PR #223, opened 2019-12-08
WIP - replace interval field with duration field

### LINKED ISSUE #222: Replace uses of Interval Library with Duration Field
The [interval-field](https://github.com/mpasternak/django-interval-field) is only supported up to Django 1.8.  Django's own duration field provides similar functionality, storing a time delta.  

I grep'ed the code base and here's the results showing it's only used on the Topics model.  

```bash
/chipy.org$ grep -R -i --include "*.py" interval .
./chipy_org/settings.py:    'interval',
./chipy_org/apps/meetings/models.py:from interval.fields import IntervalField
./chipy_org/apps/meetings/models.py:    length = IntervalField(
./chipy_org/apps/meetings/migrations/0001_initial.py:import interval.fields
./chipy_org/apps/meetings/migrations/0001_initial.py:                ('length', interval.fields.IntervalField(null=True, blank=True)),
```
The library is only really used in the Topics model and the related form.  My main concern is the preservation of data when migrating the model.  Custom migration?

elmq0022:
Closes #222 

This is a still a work in progress as I would like some feedback on design.  Mainly do we want to 

1. preserve the old duration for the talks/topics
2. have a better widget for entering the duration

I personally don't care about the old topic times all that much.  Does anyone on the boar use the historical information.  For example do  we want to know: average python talk time is 15-20 minutes, etc.?  If we want to keep the old data, I can write a custom migration.  

The default widget for this field is a text box with time specified with a format of DD HH:MM:SS.uuuuuu.  This is not the most intuitive entry as a user would need to typically type "15:00" for a 15 minute talk. Type 15 would be better.  I can either put a label with the format "MM:SS" or try to add some custom widget to make this more clear.   

My suggestions would  are:
1. forget about the old data 
2. add the time format of HH:MM:SS to the talk form; admin can correct when approving if needed
3. open a new issue to upgrade the widget if users don't follow the format

Also, I'm not sure why I ended up with migrations in the subgroups and profiles applications.

### DIFF
--- a/chipy_org/apps/meetings/migrations/0009_auto_20191208_1006.py
+++ b/chipy_org/apps/meetings/migrations/0009_auto_20191208_1006.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import tinymce.models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('meetings', '0008_auto_20190711_1501'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='meeting',
+            name='description',
+            field=tinymce.models.HTMLField(blank=True, null=True),
+        ),
+        migrations.AlterField(
+            model_name='topic',
+            name='description',
+            field=tinymce.models.HTMLField(verbose_name='Public Description', blank=True, null=True, help_text='This will be the public talk description.'),
+        ),
+        migrations.AlterField(
+            model_name='topic',
+            name='length',
+            field=models.DurationField(blank=True, null=True),
+        ),
+    ]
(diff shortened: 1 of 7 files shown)


elmq0022:
Wondering if I am heading down a reasonable path here.  Also, do we need an interval here, or will an integer suffice.  

Thanks;

raymondberg:
I think we WOULD want to preserve old data, but I'm open to that. The ways I've seen that done before usually involve a two-stage migration pattern. 

1. roll out the new field
1. cut over the values
1. switch to using exclusively the new field
1. drop the old column

elmq0022:
Thanks Ray.  That sounds like a good strategy.  What's the typical way to cut over the values?  I could write a custom management command for one time use.

elmq0022:
Also I'll dig through the code base a bit, but my guess is an integer value of minutes will suffice.  I personally don't see the need for a python timedelta if we aren't using it for any computations.

elmq0022:
close in favor of pull request #233

PR #225, opened 2019-12-14
Produce Sorted RSVP CSV File

elmq0022:
This is a  work in progress (WIP).  I would like feedback if you have it.  

I'll  clean the code a bit more and put back the CAPTCHA  before removing WIP status.


elmq0022:
This is ready to go.  The test suite passed, and I did some local user testing which worked just fine.

chipy_org/apps/meetings/views.py
```diff
@@ -200,47 +199,45 @@ def get_context_data(self, **kwargs):
...
     def _lookup_rsvps(self, rsvp):
+
+        # yield the CVS header
         if self.private:
             yield ["User Id", "Username", "Full Name",
                    "First Name", "Last Name", "Email", "Guests", ]
         else:
             yield ["Full Name", "First Name", "Last Name", "Guests", ]
+
+        data = []
```
raymondberg: I think we should switch to a different queryset rather than building an in-memory array here.

chipy_org/apps/meetings/views.py
```diff
@@ -200,47 +199,45 @@ def get_context_data(self, **kwargs):
...
+                    item.users_full_name,
+                    item.users_first_name,
+                    item.users_last_name,
+                    item.users_guests,
+                ])
+
+        # set keys for sort based on private status and
+        # use stable sort in reverse for correct multisort
+        last_name_key = 4 if self.private else 2
+        first_name_key = 3 if self.private else 1
```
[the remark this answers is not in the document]
elmq0022: That's good feedback thanks!

chipy_org/apps/meetings/views.py
elmq0022: @raymonndberg as we discussed, a better solution would be completely populating the RSVP model from anonymous user input or the logged in user model with a forced user review of their information.   Either way, the user should be able to review and update the RSVP to correct their information or modify their status.   Also I think we should make email mandatory as somewhat reasonable unique identifier.

elmq0022:
closing this in favor of #229

PR #226, opened 2019-12-16
added --user app to the make shell command

elmq0022:
I tried running the command `python manage.py makemigrations` from the docker shell which failed as I did not have rights to write the migration files to the directory.  Opening the shell as the user "app" resolves this issue.


raymondberg:
Interesting, I'd love to see this in action. Sam and I don't seem to have this issue, but we're on macs so 🤷‍♂

Issue #22, opened 2013-02-18
RSVP buttons should not show current State


brianray:
There should be no "Yes" botton if I already voted Yes.

elmq0022:
Fixed with #229

Issue #228, opened 2019-12-25
Add Test Keys and Test Secret to docker.env.test File


elmq0022:
In docker.env.sample set the norecaptcha key and secret to their test/development versions as listed in the google recaptach [faq](https://developers.google.com/recaptcha/docs/faq)

```
NORECAPTCHA_SITE_KEY=your_recaptcha_public_key
NORECAPTCHA_SECRET_KEY=your_recaptcha_private_key
```

This would allow the captcha's to work for local development and testing.

PR #229, opened 2019-12-25
RSVP Normalization

elmq0022:
Goals for this Pull Request:

1. Meeting RSVP CSVs in sorted order by Last Name, First name
1. Populate RSVP Model Completely 
1. If user is logged in populate RSVP form with user's data
1. remove "Maybe" as an RSVP option
1. remove guests from the RSVP options as meeting attendees must be on list and display a legal ID
1. move logic out front end javascript and to views / forms

Everything seems to work, but I should do some additional clean up on:
1. meeting views.py
1. meeting urls.py

Updating tests and linting pending.

Please review updates and provide feedback if possible.  Thanks!


elmq0022:
@smandekar1 happy to have your feedback as well if you have time. Thanks! 

Also see issue #228 for an easy way to add the captcha test keys to your development environment.

smandekar1:
@elmq0022  Looks good, I didn't have a chance to run it though.

elmq0022:
Thanks @smanderkar1 I've been tweaking some of the implementation to remove the JavaScript hack that I did.  It's in pretty good shape now.  

The latest commit broke the authenticated user update from home.  :sad:

I will probably fix that tommorow and remove the WIP status.

elmq0022:
close #165 (remove maybe from RSVP)

elmq0022:
close #165 (remove maybe from RSVP)
close #230 (sorted RSVPs)

Seems to work well overall.  I made a "Mixin" to get intial form data (user, meeting, etc.) used to populate the form.  Would this be better on the associated form as a static method?  

Another method would be what I implemented earlier,  GET method on the RSVP View with JavaScript used to fetch a populated form and add it to the DOM.  

**Linted and tests pass**

elmq0022:
Close #22

chipy_org/apps/meetings/urls.py
```diff
@@ -23,19 +25,18 @@
     url(r'^(?P<pk>[0-9]*)/$',
         MeetingDetail.as_view(), name='meeting'),
 
+    # pass the meeting id here!!!
```
raymondberg: I'm not sure what this comment is directing. Is this a TODO or is describing a behavior of the code?

chipy_org/apps/meetings/views.py
```diff
@@ -39,6 +42,55 @@
 logger = logging.getLogger(__name__)
 
 
+class InitialRSVPMixin():
```
raymondberg: This is a really interesting refactor. I'll admit that I'm not following all of what/why this class exists, but it makes sense that this is abstracting some of the logic around the "setup" rather than RSVP management.

elmq0022: Yeah, this is one we should discuss.  Everything works, but I don't like that not super easy to follow.  

I abstracted the setup as I needed the logic for both the home page and the meeting detail page.  They instantiate the form, but don't use a form view.  

I had an earlier commit that provided a GET method and then used JavaScript on the fontend to insert the form into the DOM and add the CAPTCHA as well.  Which I thought was hard to follow...  :frown:

Regardless, it's an improvement.  I would do much more with it until I know the direction we want to push the site   For example a list of upcoming meetings and the ability to RSVPs to each meeting listed.

elmq0022:
I need to think about the migrations for this a little more.  Also, I'm unsure how this integrates with meetup.com and don't want to have it in production until then.  Likely next month.

chipy_org/apps/profiles/migrations/0004_rsvp_migration.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+import probablepeople
+
```
elmq0022: How will this work when I kill the probablepeople? Does it need to be wrapped in try, except?

chipy_org/apps/profiles/migrations/0004_rsvp_migration.py
```diff
@@ -0,0 +1,33 @@
...
+
+def display_name_to_rsvps(apps, schema_editor):
+    UserProfile = apps.get_model('profiles', 'UserProfile')
```
elmq0022: Why can't I use `apps.get_model('profiles', 'UserProfile')` in a meetings migration?

elmq0022:
I think I need to do something with the probablepeople import in the custome migrations prior to deleting if from the lib from requirements.txt and dropping the "name" field from the RSVP.    

From what I can tell the meetup.com sync is only for the meeting and we do not pull or push and user information between the sites.  Is that correct? 

Honestly, I'd be in favor of dropping the anonymous RSVP.  Most people use it, but that's probably because it the first option.  Most everyone has a google or github account.  We could add a couple more to ease the burden further.  Just a thought.

chipy_org/apps/meetings/migrations/0012_rsvp_migration.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def parse_anonymous_rsvp_names(apps, schema_editor):
+    RSVP = apps.get_model('meetings', 'RSVP')
+    anonymous_rsvps = RSVP.objects.filter(user=None)
+    for rsvp in anonymous_rsvps:
+        if rsvp.name:
+            parsed = probablepeople.tag(rsvp.name)
+            rsvp.first_name = parsed[0].get("GivenName")
+            rsvp.last_name = parsed[0].get("Surname")
```
elmq0022: should return a space by default

chipy_org/apps/profiles/migrations/0004_rsvp_migration.py
```diff
@@ -0,0 +1,33 @@
...
+    RSVP = apps.get_model('meetings', 'RSVP')
+    users = UserProfile.objects.all()
+    for user in users:
+        parsed = probablepeople.tag(user.display_name)
+        first_name = parsed[0].get("GivenName")
+        last_name = parsed[0].get("Surname")
```
elmq0022: should return a space by default

PR #233, opened 2020-01-01
Remove Interval Field from Topics Model

elmq0022:
This is the first step in removing the interval field.

This pull request does the following:
1. Adds an integer field length_minutes to Topics model
2. Adds a data migration to populate the new length_minutes field with data from the old length field
3. Updates the admin, forms and templates to only use the new length_minutes field

There are probably two more deploys needed: 
1. Delete the length field from the model 
2. Remove the django-interval-field from the project

### DIFF
--- a/chipy_org/apps/meetings/admin.py
+++ b/chipy_org/apps/meetings/admin.py
@@ -20,6 +20,7 @@ class TopicInline(admin.StackedInline):
     filter_horizontal = ['presentors']
     readonly_fields = ['modified', 'created', ]
     extra = 0
+    exclude = ['length',]
 
 
 class TopicAdmin(admin.ModelAdmin):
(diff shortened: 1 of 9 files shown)


elmq0022:
@raymondberg could you let me know how many topics objects are in the DB at this time.  If it's really significant we may want to consider a bulk update vs processing items one at a time.  Thanks!

elmq0022:
I looked at the number of past meetings from the live site and estimate there are maybe in the neighborhood no more than a 1000 topics to update.  This should be relatively quick to update.

elmq0022:
I have two draft pull request for stages 2 and 3 an upgrade to Django 1.9, but one thing at a time :sweat_smile:

elmq0022:
@raymondberg  Thanks!

PR #245, opened 2020-01-16
Update to readme.md in 'make migrate' section

sophm256:
-Changed the section regarding 'make migrate' to suggest to users what to do if they see an error
-Also corrected small typo in original sentence " Finally, you should be able to visit your site **, but** entering the following in your url bar"

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,11 @@ up the web app and a database as services using docker-compose.
     make up
 
 After running `make up`, you need to migrate the database. This will
-create the tables and database objects needed to run the site.
+create the tables and database objects needed to run the site. 
+
+Note: After running `make migrate` the first time, an error message may appear. If the error message looks like `psycopg2.ProgrammingError: relation "django_site" does not exist`, you can ignore it. Run `make migrate` a second time. This error message should now be gone. 
+
+If after running `make migrate` the first time you get a different error message than the one above, run `make migrate` again. If the error message still appears, you should troubleshoot it.
 
     make migrate
 
@@ -50,7 +54,7 @@ Next, you should create a superuser to use to login to the site admin with.
 
     docker-compose exec web ./manage.py createsuperuser
 
-Finally, you should be able to visit your site, but entering the
+Finally, you should be able to visit your site by entering the
 following in your url bar:
 
     http://localhost:8000


README.md
```diff
@@ -42,15 +42,19 @@ up the web app and a database as services using docker-compose.
     make up
 
 After running `make up`, you need to migrate the database. This will
-create the tables and database objects needed to run the site.
+create the tables and database objects needed to run the site. 
+
+Note: After running `make migrate` the first time, an error message may appear. If the error message looks like `psycopg2.ProgrammingError: relation "django_site" does not exist`, you can ignore it. Run `make migrate` a second time. This error message should now be gone.
```
elmq0022: This is good but slightly inaccurate.  The make migrate command does calls two migrate commands

1. manage.py migrate auth 
2. manage.py migrate

And the first one does throw an error.  Then the second migration completes all remaining migrations.  (There's an intedependance between apps.) However, we should not have to call make migrate twice.  

If make migrate  is needed more than once, then there is an issue to address and hopefully some will point it out.  

Could you revise this so it's more in line with the comments above.  

Please let me know if you do indeed need to use make migrate more than once.  I have rebuilt the env from master successful several times the last two days.  

Thanks!

sophm256: Sorry, I'm not quite understanding how this works. 

> And the first one does throw an error. Then the second migration completes all remaining migrations. (There's an intedependance between apps.) However, we should not have to call make migrate twice.

So you're saying the error message is from the 1st command: `manage.py migrate auth` . I don't understand why the error disappears by the second run of `make migrate`. Does running the 2nd command: `manage.py migrate` fix the errors from the first command? If so, why would that be? Also, instead of running `make migrate` a second time, what would you suggest the user do to figure out that there is an error that persists or not?  

For example in both of my cases after running `make migrate` the first time:
 case 1. error message is  `psycopg2.ProgrammingError: relation "django_site" does not exist`
 case 2. other error message

Could you suggest what to run?  Thanks!

Also, why would it be bad or rather unnecessary to run `make migrate` a second time? I don't have a lot of experience with 'make' or Django migrations. Thanks for reviewing my request and for the suggestions!

sophm256: Thanks for the explanation. Things are much clearer now. You're right about not having to run `make migrate` a second time. How about I say:

>  Note: After executing `make migrate`, there will be an error message.  If the message reads `psycopg2.ProgrammingError: relation "django_site" does not exist`, you can ignore this error message. 
> 
> Confirm that the migrations were successful. To do this, run `make shell`, which gives you a bash shell within Docker. Then run `python manage.py migrate --list`. This shows a list of all the migrations. Make sure each migration has a marked checkbox, such as `[X] 0001_initial` . Then exit out of the shell by typing `exit` .
> 
>  If after running `make migrate` for the first time, you get a different error message than the message above, you should troubleshoot it.

Please let me know if this sounds better. Thanks!

elmq0022: That's a nice update.  Push the change on this branch and I will approve your pull request. :smile:

sophm256: Great! I have pushed the changes. Thanks for your help :)

PR #250, opened 2020-01-26
Added custom meeting titles

### LINKED ISSUE #218: Add meeting titles
Right now the only way to get a custom title for meeting is to make a new "meeting type". As a result, for the one-off jobs event, I had to create a whole meeting type just to add a title to my post. It would be nice if meeting type gave a default title, but if no meeting type was selected, there was an editable title field.

sophm256:
Fixes #218 

-Adds 'default_title' field to 'meeting_type'
-Adds 'custom_title' field to 'meeting'

If the 'custom title' field, that will be used as the event title. If 'custom title' field is blank, 'default_title' field will be used as the event title. 

-Also added tests to test the title functionality

### DIFF
--- a/chipy_org/apps/meetings/admin.py
+++ b/chipy_org/apps/meetings/admin.py
@@ -57,7 +57,7 @@ class MeetingForm(forms.ModelForm):
 
 
 class MeetingAdmin(admin.ModelAdmin):
-    list_display = ['when', 'where', 'created', 'modified', 'action', 'meeting_type']
+    list_display = [ 'title', 'meeting_type', 'when', 'where', 'created', 'modified', 'action' ]
     list_filter = ['meeting_type']
     form = MeetingForm
     inlines = [
--- a/chipy_org/apps/meetings/models.py
+++ b/chipy_org/apps/meetings/models.py
@@ -63,6 +63,7 @@ class MeetingType(CommonModel):
         'subgroups.SubGroup', blank=True, null=True,
         help_text='Optional Sub-group (i.e. SIG)')
     name = models.CharField(max_length=64)
+    default_title = models.CharField(max_length=64, null=True, blank=True)
     slug = models.SlugField(max_length=64, unique=True)
     description = tinymce_models.HTMLField(blank=True, null=True)
 
@@ -95,7 +96,11 @@ class Meeting(CommonModel):
         MeetingType, blank=True, null=True,
         help_text=("Type of meeting (i.e. SIG Meeting, "
                    "Mentorship Meeting, Startup Row, etc.). "
-                   "Leave this empty for the main meeting."))
+                   "Leave this empty for the main meeting. "
+                   ))
+    custom_title = models.CharField(max_length=64, null=True, blank=True,
+        help_text=("If you fill out this field, this 'custom_title' will show up as the title of the event." 
+                   ))
     description = tinymce_models.HTMLField(blank=True, null=True)
 
     def can_register(self):
@@ -124,7 +129,14 @@ class Meeting(CommonModel):
     def meetup_url(self):
         return "https://www.meetup.com/_ChiPy_/events/{}/".format(self.meetup_id)
 
-
+    @property 
+    def title(self):
+        if self.custom_title:
+            return self.custom_title
+        if self.meeting_type and self.meeting_type.default_title:
+            return self.meeting_type.default_title
+        return "In the Loop" # quasi default title for the main meeting
+    
 class Presentor(CommonModel):
 
     def __str__(self):
--- a/chipy_org/apps/meetings/templates/meetings/meeting.html
+++ b/chipy_org/apps/meetings/templates/meetings/meeting.html
@@ -13,11 +13,12 @@
 
     <div class="row-fluid">
       <div class="span12">
-          {% if meeting.meeting_type %}
-          <h3 itemprop="name">{{ meeting.meeting_type.name }}</h3>
-          {% if meeting.meeting_type.subgroup %}<p><a href="{% url 'groups' meeting.meeting_type.subgroup.slug %}">{{ meeting.meeting_type.subgroup.name }}</a></p>{% endif %}
+          {% if meeting.title %}
+              <h3 itemprop="name">{{ meeting.title }}</h3>
           {% endif %}
+
           <hr />
+
           {% if meeting.description %}
           <p itemprop="description">{{ meeting.description|bleach|safe }}</p>
           {% endif %}
--- a/chipy_org/templates/homepage.html
+++ b/chipy_org/templates/homepage.html
@@ -132,7 +132,10 @@ h2 a {
 
         <div class="row-fluid">
             <div class="module span6">
-                <h3>In the Loop</h3>
+                {% if next_meeting.title %}
+                    <h3>{{ next_meeting.title }}</h3>
+                {% endif %}
+
                 <p>Meetings happen in the downtown area the second Thursday of every month.</p>
 
                 {% if next_meeting %}
@@ -250,14 +253,13 @@ h2 a {
         <div class="row-fluid">
             {% for meeting in other_meetings %}
             <div class="module span4 box-feature" itemscope itemtype="http://schema.org/Event">
-
-                {% if meeting.meeting_type %}
-                <h4 itemprop="name">{{ meeting.meeting_type.name }}</h4>
-
-                {% if meeting.meeting_type.subgroup %}<p><a href="{% url 'groups' meeting.meeting_type.subgroup.slug %}">{{ meeting.meeting_type.subgroup.name }}</a></p>{% endif %}
-
+                
+                {% if meeting.title %}
+                    <h4 itemprop="name">{{ meeting.title }}</h4>
                 {% endif %}
+
                 <hr />
+
                 {% if meeting.description %}
                 <p itemprop="description">{{ meeting.description|truncatewords_html:30|safe }}</p>
                 {% endif %}
--- a/chipy_org/apps/meetings/migrations/0015_auto_20200208_1904.py
+++ b/chipy_org/apps/meetings/migrations/0015_auto_20200208_1904.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('meetings', '0014_merge'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='meeting',
+            name='custom_title',
+            field=models.CharField(max_length=64, blank=True, null=True, help_text="If you fill out this field, this 'custom_title' will show up as the title of the event."),
+        ),
+        migrations.AddField(
+            model_name='meetingtype',
+            name='default_title',
+            field=models.CharField(max_length=64, blank=True, null=True),
+        ),
+        migrations.AlterField(
+            model_name='meeting',
+            name='meeting_type',
+            field=models.ForeignKey(blank=True, null=True, help_text='Type of meeting (i.e. SIG Meeting, Mentorship Meeting, Startup Row, etc.). Leave this empty for the main meeting. ', to='meetings.MeetingType'),
+        ),
+    ]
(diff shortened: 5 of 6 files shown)


chipy_org/apps/meetings/admin.py
```diff
@@ -57,7 +57,7 @@ class Meta:
 
 
 class MeetingAdmin(admin.ModelAdmin):
-    list_display = ['when', 'where', 'created', 'modified', 'action', 'meeting_type']
+    list_display = ['when', 'where', 'created', 'modified', 'action', 'meeting_type', 'meeting_title']
     list_filter = ['meeting_type']
```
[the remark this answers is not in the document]
elmq0022: @sophm256 could I ask you to take a slightly different approach with this?   As implemented here the use of meeting title and meeting subgroup are getting mixed which causes several if-thens to be used in the templates.  

Instead could you add a `defaul_title` field to the meeting subgroup object?  Then in the template use the meeting title if it's available or fall back to the subgroup default title if available?

Tests would be appreciated (not required) and a good exercise if you haven't done one before.  The pytest style of a function starting with `def test_` is preferred.  Look at my recent bug fix for the meeting RSVP emails if you want an example.  

Keep up the great work.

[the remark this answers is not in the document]
sophm256: @elmq0022 Thanks for your feedback. **Can you say a bit more about what you mean when you say**

>  As implemented here the use of meeting title and meeting subgroup are getting mixed which causes several if-thens to be used in the templates.

So for the main meeting, the new code doesn't affect the main meeting.
sample-1

Here is what the new code looks like for 3 cases:
sample-2

Case 1 (Left box) : The  'Algo Sig' is chosen as a 'meeting type' from the drop down menu
Case 2 (middle box): 'Meeting-Title' is chosen from drop down menu 'meeting type'. Then 'meeting title' field is filled with "Future of Robotics"
Case 3 (right box): 'Recursive Sig' is chosen as 'meeting type'. 'My 2nd sub group' is linked to the meeting type 'Recursive Sig'.

I use elif clauses in the template so only the regular meeting types show up or a custom meeting title shows up.  It's an 'exclusive or' relationship.

> Instead could you add a defaul_title field to the meeting subgroup object? Then in the template use the meeting title if it's available or fall back to the subgroup default title if available?

meeting

If you look at the pic above, the 'meeting type' looks like 'Sig Meeting' or 'Mentorship Meeting'. This is what usually shows up as the title of the event. The meeting subgroup is an optional object and sometimes doesn't show up on the template/homepage. This is because a meeting subgroup is an optional object to be associated with the meeting type.  In my new code and the original code, the meeting subgroup's name shows up as a subtitle to the 'meeting type' when the 'meeting type' is the title of the event (which was case 3 in the picture earlier). (Below are pictures of a subgroup and a meeting type in the admin panel.)

meeting-type2
subgroup

I'm cool with implementing it how you want me to, but I wasn't sure if we were in agreement with what was going on. Then again, I may be the one that's confused :) 

Since you have access to the database, could you please give me a concrete example of what a 'meeting type' name is and what a 'subgroup' name is? Thanks. I'm unclear what is the difference between the two.

I appreciate your help and guidance on how to improve things!

elmq0022:
Let me take a little closer look at the relationship between meeting, meeting type, and subgroup.  And I will come back to you.  

My main concern is putting logic in templates.  I like my templates to be "dumb" for lack of a better term.  

Ideally, in a template we want to reference meeting.title and get the right thing. That's more intuitive in my mind.

elmq0022:
Okay, I said subgroup when I should have said meeting_type object.  My bad... 

But looking through this again, the reliance on a specific meeting type for to allow a custom title is awkward and dangerous if said type gets deleted.  Also, why can't all meetings of all meeting types have a custom title?  

Here's what I recommended for the models, and then I leave the rest to you.  

### Meeting Type
1. Add a **field** for `default_title`

Giving a meeting type will give a default title specific to that type

### Meeting Model
1. Add a **field** for `custom_title`
2. Add a **property** called `title` 

Any meeting can have a custom title and we will use a property to prefer the  custom_title or fall back to the default title if available. 

Logic for the title property should be something like this:
``` python
@property
 def title(self):
    if self.custom_title:
        return self.custom_tile
    if self.meeting_type and self.meeting_type.default_title:
        return self.meeting_type.default_title
    return '' # return an empty string if neither title is defined 
```

There will be some hiccups along the way, but this should get you going in a better direction.

sophm256:
>  the reliance on a specific meeting type for to allow a custom title is awkward and dangerous if said type gets deleted. 

Yes, I agree with this. The downside to my approach is that the user has to add a special meeting type and that meeting type can be accidentally deleted or modified. I went with this approach because it seemed to extend what code was already there in the template without changing existing functionality. And I was unsure of how to approach this. Your solution using a property is a more elegant approach.

> list_display = ['when', 'where', 'created', 'modified', 'action', 'meeting_type', 'meeting_title']
> Do these show up in order in the admin? If so could we add it as the first item instead of the last?

Yes, they show up in order. I won't add in the 'meeting_title' to the list. Instead, I'll just list 'title' first.

sophm256:
I've changed the property code slightly to account for the case that a user doesn't put in a default title.
```
@property 
    def title(self):
        if self.custom_title:
            return self.custom_title
        if self.meeting_type:
            if self.meeting_type.default_title:
                return self.meeting_type.default_title
            else:
                return self.meeting_type
        return "" # This is the case for the Main meeting
```
Also, I had a question about use case. 

Let's say there are the following meeting_type in the database:
- none (Which is currently used for Main meeting when you have no meeting type selected)
-  Algo Sig

The meeting_type 'Algo Sig' has a 'default_title' of 'Algo Sig Default Title'.
'none' can't have a 'default_title' because it's not an object. It's null in the database.

Case 1:
I want a custom title.  I enter into the field 'custom_title' the words 'My Custom Title'. I choose 'meeting_type' of 'Algo Sig'. 'My Custom Title' shows up as the title of the event instead of 'Algo Sig Default Title in the 'Upcoming Events' Section on homepage.html .  This is so far so good.

Case 2:
I want a custom title and enter into the 'custom_title' field the words  'My Custom Title'. I choose 'none' or (----) in the drop down menu 'meeting_type'. Now 'My Custom Title' will show up as the title of the event instead of the phrase 'In the Loop' in the 'Next Main Meeting' section. Ok, this is good too.

Case 3:
I want a custom title and enter into the 'custom_title' field the words  'My Custom Title'. But I do NOT want to associate it with a particular meeting type. What happens now? 

I think this may be part of the reason, I wanted to add the 'meeting_type' of 'Meeting-Type' before so it could be clear that you are want a custom title for a one off event. Because as it is, the user has to choose either 'none' or one of the existing meeting types from the drop down menu.  This because 'meeting-type' field is a required field in the 'meeting' object.

If they choose 'none' currently, their meeting will show up in the 'Next Main Meeting' section on the homepage. If they choose 'Algo Sig', their meeting will show up in the 'Upcoming Events' section on the homepage. 

Please advise what should happen in case 3. Thanks!

elmq0022:
Let's work under the assumption that if a custom title is entered that's
what the user wants regardless of any other entries.

I would not use the meeting type as a meeting title.  we have very few
meeting types. the default titles can be updated manually in the database.

elmq0022:
keep up the good work.  looks like you are making progress.

chipy_org/apps/meetings/models.py
```diff
@@ -96,7 +97,13 @@ def __str__(self):
         MeetingType, blank=True, null=True,
         help_text=("Type of meeting (i.e. SIG Meeting, "
                    "Mentorship Meeting, Startup Row, etc.). "
-                   "Leave this empty for the main meeting."))
+                   "Leave this empty for the main meeting. "
+                   ))
+    custom_title = models.CharField(max_length=64, null=True, blank=True,
+        help_text=("If you fill out this field, this 'custom_title' will show up as the title of the event."
+                   "To have the custom title show up in the Main Meeting on the homepage, select '--' from 'meeting_type' above."
+                   "To have the custom title show up in Upcoming Events on the homepage, select any other 'meeting_type'."
```
elmq0022: I think we can delete these two lines.  Our admins understand that no meeting equals main meeting.  It's hacky; we should open an issue to change that.  I would want Ray to discuss that with the user first.

chipy_org/apps/meetings/models.py
```diff
@@ -125,7 +132,14 @@ def get_absolute_url(self):
...
 
-
+    @property 
+    def title(self):
+        if self.custom_title:
+            return self.custom_title
+        if self.meeting_type and self.meeting_type.default_title:
+            return self.meeting_type.default_title
+        return "" # return an empty string if neither title is defined
+
```
elmq0022: If you want, you could make this final return the default for the main meeting as not custom title and no meeting type is a main meeting with a default title.  I'd put whatever is in the template now.

chipy_org/apps/meetings/templates/meetings/meeting.html
```diff
@@ -13,11 +13,12 @@
...
       <div class="span12">
-          {% if meeting.meeting_type %}
-          <h3 itemprop="name">{{ meeting.meeting_type.name }}</h3>
-          {% if meeting.meeting_type.subgroup %}<p><a href="{% url 'groups' meeting.meeting_type.subgroup.slug %}">{{ meeting.meeting_type.subgroup.name }}</a></p>{% endif %}
+          {% if meeting.title %}
+              <h3 itemprop="name">{{ meeting.title }}</h3>
           {% endif %}
+
           <hr />
+
```
elmq0022: That's much cleaner and clearer than the previous pull request.  Very nice improvement.  :smile:

chipy_org/templates/homepage.html
```diff
@@ -132,7 +132,12 @@ <h4>{{ announcement.headline }}</h4>
 
         <div class="row-fluid">
             <div class="module span6">
-                <h3>In the Loop</h3>
+                {% if next_meeting.custom_title %}
+                    <h3>{{ next_meeting.custom_title }}</h3>
+                {% else %}        
+                    <h3>In the Loop</h3>
+                {% endif %}
```
elmq0022: I don't see any reason why this wouldn't work, but why not just use `next_meeting.title`?

sophm256: next_meeting refers to when meeting_type is null. Since it is null and not an object, it can not have a 'title' field.  

If you want to say 'next_meeting.title', the user would have to make a 'meeting_type' named 'main meeting'. Then I could add a 'title' field to that object.  I'd also have to change the code in `/chip_org/apps/meetings/views.py` under the InitialRSVPMixin class.

Currently the code is: 

https://github.com/chicagopython/chipy.org/blob/master/chipy_org/apps/meetings/views.py#L45
```
class InitialRSVPMixin():
    def get_next_main_meeting(self):
        return (Meeting.objects
                .filter(meeting_type__isnull=True)
                .filter(when__gt=datetime.datetime.now()-datetime.timedelta(hours=6))
                .order_by('when')
                .first())
```

sophm256:
Thanks for your help! I agree, the code looks much cleaner.

chipy_org/apps/meetings/models.py
sophm256: Sorry, could you please explain this? I'm not following.

> If you want, you could make this final return the default for the main meeting as not custom title and no meeting type is a main meeting with a default title. I'd put whatever is in the template now.

chipy_org/templates/homepage.html
sophm256: Ok, I just read your comment from way above. I wasn't reading your comments in order! 

> If you want, you could make this final return the default for the main meeting as not custom title and no meeting type is a main meeting with a default title. I'd put whatever is in the template now.

So if I understand your comment correctly, we could define a default title for the main meeting here. 
> return "" # return an empty string if neither title is defined

And then yes, I can change the template to be 'next_meeting.title' without having to change any other code.

sophm256: Originally, I wrote the code that way because next_meeting in this case refers to a meeting_type of null. And null can't have a 'default_title' since it isn't an object. 

https://github.com/chicagopython/chipy.org/blob/master/chipy_org/apps/meetings/views.py

```class InitialRSVPMixin():
    def get_next_main_meeting(self):
        return (Meeting.objects
                .filter(meeting_type__isnull=True)
                .filter(when__gt=datetime.datetime.now()-datetime.timedelta(hours=6))
                .order_by('when')
                .first())
```
Now with your suggestion, I can just define the 'default_title' as 'In the Loop'.  Nice catch!

sophm256:
Made two commits. In the first, I made the changes you asked for. In the second, I wrote a few tests.

raymondberg:
I'd like to squash this down and redo the migrations. I'm happy to pair on this remotely sometime soon.

PR #254, opened 2020-02-09
Add resetdb

raymondberg:
## Problem

Resetting a database when migrations are involved is a pain. It should be really easy to get back to a fresh db image.

## Solution

Add `make resetdb`. This will drop the main schema (`public`) and all of its tables. Once ready, it will recreate the main schema.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,10 @@ shell:
 psql:
 	@docker-compose exec db psql chipy chipy
 
+resetdb:
+	@docker-compose exec db psql chipy chipy -c "drop schema if exists public cascade;"
+	@docker-compose exec db psql chipy chipy -c "create schema public;"
+
 web: run
 
 migrate:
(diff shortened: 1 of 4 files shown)


elmq0022:
Actually, do we really want this?  We should be careful to run Migrations from a starter that's similar to prod to ensure we don't run into issues during deploy.  Also makes it easier to accidentally drop the DB in prod when trouble shooting.  

What would be really great is it we could pass the -p flag to all our docker-compose commands.

raymondberg:
Docker compose commands wouldn't work in prod because it's all on Heroku.

My thought on this is that it is truly a reset db. It should wipe out what we have and start fresh. This was a big problem when working with some of the PRs where people were migrating. Folks shouldn't need to know about docker volumes, ideally. This should just get the DB to a clean state.

elmq0022:
Sounds good 👍.  Thanks for the additional color.

PR #262, opened 2020-03-09
Standardize formatting

raymondberg:
## Problem

We have a lot of legacy string formatting across the repo. 

## Solution

Simple f-string updates where possible. There's second commit to ditch the `str % params` format in one place where we use localized text elements.

@chicagopython/web-guild

### DIFF
--- a/chipy_org/apps/meetings/models.py
+++ b/chipy_org/apps/meetings/models.py
@@ -43,7 +43,7 @@ class Venue(CommonModel):
         Use the string returned as args for google.maps.LatLng constructor.
         '''
         if self.latitude is not None and self.longitude is not None:
-            return "%.6f,%.6f" % (self.latitude, self.longitude)
+            return f"{self.latitude:.6},{self.longitude:.6}"
 
     directions = models.TextField(blank=True, null=True)
     embed_map = models.TextField(blank=True, null=True)
@@ -68,7 +68,7 @@ class MeetingType(CommonModel):
     description = tinymce_models.HTMLField(blank=True, null=True)
 
     def __str__(self):
-        return "%s | (%s)" % (self.id, self.name)
+        return f"{self.id} | ({self.name})"
 
     class Meta(object):
         verbose_name = "Meeting Type"
@@ -79,9 +79,8 @@ class Meeting(CommonModel):
 
     def __str__(self):
         if self.where:
-            return "%s at %s" % (
-                self.when.strftime("%a, %b %d %Y at %I:%M %p"), self.where.name)
-        return "%s location TBD" % self.when
+            return "{self.when:%a, %b %d %Y at %I:%M %p} at {self.where.name}"
+        return f"{self.when} location TBD"
 
     when = models.DateTimeField()
     reg_close_date = models.DateTimeField(
@@ -129,7 +128,7 @@ class Meeting(CommonModel):
         return reverse("meeting", args=[self.id])
 
     def meetup_url(self):
-        return "https://www.meetup.com/_ChiPy_/events/{}/".format(self.meetup_id)
+        return f"https://www.meetup.com/_ChiPy_/events/{self.meetup_id}/"
 
     @property
     def title(self):
@@ -142,7 +141,7 @@ class Meeting(CommonModel):
 class Presentor(CommonModel):
 
     def __str__(self):
-        return "%s | (%s)" % (self.name, self.email)
+        return f"{self.name} | ({self.email})"
 
     user = models.ForeignKey(User, blank=True, null=True)
     name = models.CharField(max_length=MAX_LENGTH)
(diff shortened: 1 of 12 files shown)


chipy_org/libs/social_auth_pipelines.py
```diff
@@ -18,6 +18,4 @@ def associate_by_email(*args, **kwargs):
         if User.objects.filter(email=email).exists():
             msg = ugettext('This email is already in use. First login with your other account and '
                            'under the top right menu click add account.')
-            raise AuthAlreadyAssociated(backend, msg % {
-                'provider': backend.name
-            })
+            raise AuthAlreadyAssociated(backend, msg.format(provider=backend.name))
```
elmq0022: I don't see either one of these formatted strings doing anything besides return the string in the ugettext argument.   Maybe the backend was supposed to be provided.  Maybe we could add it now?  

Also, why do we use `ugettext` here? Isn't that for translation? Do we do translation for chipy.org?

chipy_org/apps/meetings/models.py
```diff
@@ -79,9 +79,8 @@ class Meeting(CommonModel):
 
     def __str__(self):
         if self.where:
-            return "%s at %s" % (
-                self.when.strftime("%a, %b %d %Y at %I:%M %p"), self.where.name)
-        return "%s location TBD" % self.when
+            return "{self.when:%a, %b %d %Y at %I:%M %p} at {self.where.name}"
+        return f"{self.when} location TBD"
```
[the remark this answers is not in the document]
sophm256: Yes, the f is missing on line 82.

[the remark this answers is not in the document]
sophm256: @raymondberg , but @elmq0022's suggestion had a small typo. The f should come before the first quotation mark, not after.  So line 82 should be:

 `return f"{self.when:%a, %b %d %Y at %I:%M %p} at {self.where.name}"`

chipy_org/libs/social_auth_pipelines.py
sophm256: In the code's current format, the string representation of the 'backend' object will show up in the raise error message since 'backend' is in the parentheses.  This 'backend' object string will be concatenated with the variable 'msg' since there's a comma between them.

A raise statement in Python is usually in the form of:
 `raise SomeError("some error string")`

What will not show up in our raised error message is 'backend.name' specifically. This is because 'msg' does not have 'backend.name' in it.

So in line with @elmq0022's  suggestions, maybe the solution might look like this:

```
msg = f"Backend is {backend}. Backend name is {backend.name}. This email from is already in use. First login with your other account and under the top right menu click add account."

raise AuthAlreadyAssociated(msg)
```

@raymondberg Feel free to change it however you see fit.

chipy_org/apps/meetings/models.py
```diff
@@ -43,7 +43,7 @@ def jsonLatLng(self):  # pylint: disable=invalid-name
         Use the string returned as args for google.maps.LatLng constructor.
         '''
         if self.latitude is not None and self.longitude is not None:
-            return "%.6f,%.6f" % (self.latitude, self.longitude)
+            return f"{self.latitude:.6},{self.longitude:.6}"
```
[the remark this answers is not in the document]
sophm256: @raymondberg, @elmq0022 Yes, I agree that it should be: `f"{self.latitude:.6f},{self.longitude:.6f}"`

Please see https://stackoverflow.com/questions/45310254/fixed-digits-after-decimal-with-f-strings for more of a discussion on this.

sophm256:
Overall, great job @raymondberg . You're a champ for looking through and revising all of these strings.

sophm256:
And great job @elmq0022 for the well thought out suggestions!

chipy_org/libs/social_auth_pipelines.py
```diff
@@ -16,6 +16,8 @@ def associate_by_email(*args, **kwargs):
     if email:
         User = get_user_model()
         if User.objects.filter(email=email).exists():
-            msg = ugettext('This email is already in use. First login with your other account and '
-                           'under the top right menu click add account.')
-            raise AuthAlreadyAssociated(backend, msg.format(provider=backend.name))
+            msg = (f"Backend is {backend}. Backend name is {backend.name}."
+                    "This email from is already in use. First login with your "
+                    "other account and under the top right menu click add account.")
```
elmq0022: I can't find a __str__ or __rep__ implementation associated with the Backend or and of its inherited classes, so I don't know what it will look like.  I would just remove the `f"Backend is {backend}. Backend name is {backend.name}."` portion of this.  

It would be more helpful if we could say "this email is already associate with the {backend}, please login with that account first and then associate this account with it"

That's a bigger lift and would skip for this pull request. 

I'm approving, but would prefer this change to be made before deploy.  If not, we'll get it in the next one.  

Thanks!

raymondberg: I compromised and wrote something a little clearer using just `backend.name`.

elmq0022:
That's cool.  Thanks for letting me know!

Issue #264, opened 2020-03-10
ChiPy Announcement Dialog in the admin is not editable


JoeJasinski:
On the announcement detail page, the announcement "Text" field is grayed out. 

https://www.chipy.org/admin/announcements/announcement/58/change/

![image](https://user-images.githubusercontent.com/417872/76368382-1282d200-62fe-11ea-959e-7fad17543a44.png)

elmq0022:
Thanks Joe.  I will look into it as soon as I can.

elmq0022:
@JoeJasinski  I took a look my local dev with the deployed code.  and a deploy to heroku.  Both work just fine. This unfortunately makes it hard for me to know where to start to fix the issue.    

![screen](https://user-images.githubusercontent.com/3450848/76374773-1f112580-6312-11ea-96cd-fbc1d2014039.png)
@raymondberg my credentials don't give me access to announcements on the live site.  Could you verify @JoeJasinski 's issue.  Could you try refreshing the cache if it doesn't?  

On a separate note, we have two "rich" text editor on the site.  How would you guys feel about ditching one of them?  The one that's causing problems right now is ckeditor.  The other is tinymce (sp?).  

Could we swap the ckeditor for tinymce? 

Thanks!

elmq0022:
@JoeJasinski I see you did get an announcement up today.  Were you able to get it to work through the admin, or did you need to use a some backdoor?  

I also see the prior announcement was posted December 16th, which is actually several deploys back.  I'll see if I can go back to a previous deploy and find where it broke.  It might be easier to try deploying an update version though and seeing if that resolves it.

sophm256:
@elmq0022 I'm just going to throw out some ideas which might or might not work. 

idea 1) do a collectstatic in production  

https://stackoverflow.com/questions/57089999/ckeditor-in-django-not-working-in-production 

https://stackoverflow.com/questions/49020840/django-2-0-1-with-ckeditor-doesnt-work-on-admin-page/54272536#54272536

idea 2) That second link has an answer for django 2, while I know we've upgraded to django 3. The first and most upvoted answer has to do with specifying a path to find the editor in a static directory. We currently don't have this defined in our settings.

idea 3) Two of the answers in 2nd link have to do with static files. Maybe S3 settings are not optimal?

idea 4) Install ckeditor without extra 3rd party plugins and work from there. The 2 links are solutions for the toolbar being greyed out, whereas we have the problem of them not showing up. But it may be worth a shot.
[...]

PR #265, opened 2020-03-11
Upgrade django storages

elmq0022:
This will put the sponsor logos back on the site.  Old version of django-storages is  not compatible with python3.


elmq0022:
Added an upgrade to django-ckeditor to 5.9.0.  Hoping this resolves issue #264, but it's kind of a shot in the dark.

elmq0022:
Also the boto3 upgrade should fix our failing circleci test.

PR #266, opened 2020-03-21
Test fixes

raymondberg:
Hey, our tests are broken! 

We really need to figure out how to not serve HTTPS in CI, but I kinda just want the tests to pass. This does it.

@sophm256 and I trudged our way through this one :D

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,7 @@ jobs:
     docker:
       - image: circleci/python:3.6.9
         environment:
+          ALLOWED_HOSTS: localhost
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     docker:
       - image: circleci/python:3.6.9
         environment:
+          ALLOWED_HOSTS: localhost
+          CONNECT_METHOD: http
```
raymondberg: Truly, we don't need these, but I think we should have them match what's in docker-compose.yaml until we know better.

PR #267, opened 2020-03-23
Homepage logo

sophm256:
This is still a work in progress. The original homepage hardcoded "Platinum Sponsors". I changed the homepage template so that the sponsor groups would be listed from the database. I attempted to code some logic into the template so that each sponsor group shows up once, but I can't figure out how to do that.  For example, in my setup below "silver" shows up twice because there are two sponsors in the silver group. Currently, each sponsor group shows up before each sponsor. 

Do you have any suggestions on how to do this? I know it's not great to put logic in templates, but there is no easy way to do this. I tried to instill some logic by setting a variable using 'with' outside of the 'for' loop and then I wanted to reassign that same variable in the 'for' loop, but the computer complained and had me put 'endswith' in several places.

Also, I have coded it so that even if a sponsor has no logo, their name still shows up.

Below is my test setup:
Platinum group, sponsor1 has no logo
Gold group, sponsor2  has an orange logo
Silver group,  sponsor3 has no logo
Silver group, sponsor4 has a blue logo

homepage-logos


sophm256:
The other way to have only a sponsor group show up once would be to have a list in the template.
Here's pseudocode:
```
my_list = []
if sponsor_group not in my_list:
    my_list.append(sponsor_group)
    print(sponsor_group)
```

elmq0022:
@sophm256 maybe you could look at this tag [ifchanged](https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#ifchanged) 

I would have something like: 

```python
[(sponsor1, group1), ..., (sponsor_n, group_n)]
```

and loop through this and print the header if the group changes when it changes.

I used tuples in the example above, but you could just as well use the model object so something like this to override the model in the list view

```python
s = Sponsors.objects.all().groupby('sponsor_group')
```
I see what you are doing in the approach above, and you can definitely make it work, but consider what happens when a new sponsor group is added. :wink:

You've got a good start.  Let's get you another pull request. 

Go head an reopen this if it makes sense to you.

sophm256:
@elmq0022 Thanks for this!  The 'ifchanged' tag did the trick and was what I was looking for. ☺️ 

 In the homepage template, I pretty much kept everything else the same and added the 'ifchanged' tag.  So this implementation will use general_sponsors after all. I also changed the ordering of sponsor/general_sponsors by name ascending for both the homepage and sponsor list page.

elmq0022:
Great! Happy to see you getting another pull request.

chipy_org/apps/sponsors/models.py
```diff
@@ -61,3 +61,6 @@ def __str__(self):
 
     def get_absolute_url(self):
         return reverse("sponsor_detail", args=[self.slug])
+
+    class Meta:
+        ordering = ["name"]
```
elmq0022, line 66: from pylint:

************* Module chipy_org.apps.sponsors.models
C: 66, 0: Final newline missing (missing-final-newline)
C: 66, 0: Mixed line endings LF and CRLF (mixed-line-endings)

chipy_org/apps/main/views.py
```diff
@@ -32,7 +32,7 @@ def get_context_data(self, **kwargs):
         context = {}
         context.update(kwargs)
         context['other_meetings'] = self.get_non_main_meetings(num=3)
-        context["general_sponsors"] = GeneralSponsor.objects.all().order_by('?')
+        context["general_sponsors"] = GeneralSponsor.objects.all().order_by('sponsor__sponsor_group__list_priority', 'sponsor__name')
         context['announcement'] = Announcement.objects.featured()
```
elmq0022, line 36: from pylint: 

************* Module chipy_org.apps.main.views
C: 35, 0: Line too long (133/100) (line-too-long)

sophm256:
Thanks for taking a look at this pull request! Should I try to make any changes to address the linting errors? I'm not even sure how I'm getting a mixed LF and CRLF for line 66 because my environment for editing that code is on Ubuntu.

elmq0022:
If you have time ahead of Thursday, yes please resolve the lint issues.

sophm256:
Linting issues have been fixed.

PR #269, opened 2020-03-26
Automatic Meeting Scheduler

sophm256:
This what I've done so far. I've created the RepeatMeetings model, which is a copy of the Meetings model but modified. I also modified the admin.py file. I created 'RepeatMeetingAdmin' which is a copy of 'MeetingAdmin' but modified. I deleted `ordering = ('-when',)` in admin.py because there was no longer a 'when' field. In 'RepeatMeetingAdmin' also had to comment out the lines
```
inlines = [
     TopicInline,
     MeetingSponsorInline,
    ]
```
because I was getting some errors.

Before I commented out those lines I got this error:
meetingSponsor-has-no-ForeignKey-error

Then when I commented out 'TopicInline' from that 'inlines' variable, I got this error:
Topic-has-no-foreignkey-error

So what do you think I should do about these errors? (I mean they don't show up once I comment out the whole 'inlines' variable.) Is it alright to remove them permanently?


sophm256:
I'm going to close this PR and will start over with new one in the future whenever I get around to working on it. This is because master has extensively changed in the last deployment.

PR #272, opened 2020-04-14
added the presenters to the topic admin list page

JoeJasinski:
As a speaker coordinator, I want to view what speakers are associated with what topics as I browse the topics list. 

This PR is to make it easier to see who is associated with topics that are submitted without drilling down into the topic detail page. 

Screen Shot 2020-04-13 at 9 37 29 PM

This PR also:
 - Adds a name field to the Speaker List page
 - Fixes a recursive call bug with the display of Topics when the adding a topic for the first time and no presenters are attached.

### DIFF
--- a/chipy_org/apps/meetings/admin.py
+++ b/chipy_org/apps/meetings/admin.py
@@ -23,7 +23,7 @@ class TopicInline(admin.StackedInline):
 
 
 class TopicAdmin(admin.ModelAdmin):
-    list_display = ('id', 'approved', 'title', 'experience_level', 'meeting', 'created')
+    list_display = ('id', 'approved', 'title', 'experience_level', 'get_presenters', 'meeting', 'created')
     readonly_fields = ['get_presenters', 'modified', 'created', ]
     list_filter = ['approved', 'experience_level']
     search_fields = ['title']
@@ -31,7 +31,7 @@ class TopicAdmin(admin.ModelAdmin):
 
     def get_presenters(self, obj):
         return format_html(" &bull; ".join(
-            [f"<a href='{reverse('admin:meetings_presentor_change', args=[p.id])}'>{p}</a>"
+            [f"<a href='{reverse('admin:meetings_presentor_change', args=[p.id])}'>{p.name}</a>"
              for p
              in obj.presentors.all()]))
     get_presenters.short_description = "Presenter Links"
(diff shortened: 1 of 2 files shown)


chipy_org/apps/meetings/models.py
```diff
@@ -178,8 +178,6 @@ class Topic(CommonModel):
 
     def __str__(self):
         out = self.title
-        if self.presentors.count():
```
JoeJasinski, line 181: This was causing a recursive call when adding a new topic. We may want to rethink this logic. Removing for now to unblock the page.

chipy_org/apps/meetings/admin.py
```diff
@@ -79,7 +79,7 @@ class Media:
 
 class PresentorAdmin(admin.ModelAdmin):
     list_display = [
-        'id', 'user', 'email', 'phone']
+        'id', 'name', 'user', 'email', 'phone']
```
JoeJasinski, line 82: Adding name field to the Presenter List, as it's possible that this is the only field filled out by a presenter and the list page would contain an empty looking record otherwise.

chipy_org/apps/meetings/admin.py
```diff
@@ -23,15 +23,15 @@ class TopicInline(admin.StackedInline):
...
+    list_display = ('id', 'approved', 'title', 'experience_level', 'get_presenters', 'meeting', 'created')
     readonly_fields = ['get_presenters', 'modified', 'created', ]
     list_filter = ['approved', 'experience_level']
     search_fields = ['title']
     filter_horizontal = ['presentors']
 
     def get_presenters(self, obj):
         return format_html(" &bull; ".join(
-            [f"<a href='{reverse('admin:meetings_presentor_change', args=[p.id])}'>{p}</a>"
+            [f"<a href='{reverse('admin:meetings_presentor_change', args=[p.id])}'>{p.name}</a>"
```
JoeJasinski, line 34: This is so we limit to just seeing the name and not the markup that's part of __str__

chipy_org/apps/meetings/admin.py
```diff
@@ -23,15 +23,15 @@ class TopicInline(admin.StackedInline):
 
 
 class TopicAdmin(admin.ModelAdmin):
-    list_display = ('id', 'approved', 'title', 'experience_level', 'meeting', 'created')
+    list_display = ('id', 'approved', 'title', 'experience_level', 'get_presenters', 'meeting', 'created')
```
JoeJasinski, line 26: This is the core of the change. See PR description.

PR #277, opened 2020-04-18
Django 2.2 upgrade

elmq0022:
Upgrades to Django 2.2.12 and Python 3.8.2.  

I'm leaving this in draft as it needs some testing and lint fixes.


elmq0022:
Testing Progress (mostly notes for myself)

1. The admin pops (the little plus sign for related fields) are not popping up.  The spawn a new window.  This is not the case with the current deploy on chipy.org.  
2. Displaying "My Talks" is throwing an error.  Again production is just fine.   
screen

screen

elmq0022:
Looks like we move from jquery 2.2.3 to jquery 3.3.1 in this release.  There are several breaking changes in this version around the ajax method and cross domain scripting.  

[jquery upgrade](https://jquery.com/upgrade-guide/3.0/)

elmq0022:
There's some minor lints to fix with the upgrade to pylint, but the test suite passes and the major functionality all seems to be in order.

chipy_org/apps/meetings/admin.py
```diff
@@ -48,7 +48,7 @@ def get_presenters(self, obj):
             " &bull; ".join(
                 [
                     f"<a href='{reverse('admin:meetings_presentor_change', args=[p.id])}'>"
-                    "{p.name}</a>"
+                    f"{p.name}</a>"
```
raymondberg, line 51: Fun

chipy_org/settings.py
```diff
@@ -156,7 +156,6 @@ def env_list(key, defaults=None, delimiter=","):
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
-    "django.contrib.auth.middleware.SessionAuthenticationMiddleware",
```
raymondberg, line 159: In case you, like me, were wondering: https://docs.djangoproject.com/en/3.0/releases/2.0/#miscellaneous

elmq0022: Yeah, you can imagine my surprise when that threw an error.  😉

-

chipy_org/apps/sponsors/models.py
```diff
@@ -21,8 +22,8 @@ class Meta:
 
 
 class GeneralSponsor(models.Model):
-    sponsor = models.ForeignKey("sponsors.Sponsor")
-    about = models.TextField("About this sponsorship", blank=True, null=True)
+    sponsor = models.ForeignKey("sponsors.Sponsor", on_delete=models.CASCADE)
+    about = models.TextField(u"About this sponsorship", blank=True, null=True)
     about_short = models.CharField(
         "Brief description of sponsorship", max_length=128, blank=True, null=True
```
sophm256: A similar thing to the above comment. The 'about' TextField here is a unicode string, while the 'about_short' CharField is not.

chipy_org/apps/sponsors/models.py
```diff
@@ -4,9 +4,10 @@
...
 class MeetingSponsor(models.Model):
-    sponsor = models.ForeignKey("sponsors.Sponsor")
-    meeting = models.ForeignKey("meetings.Meeting", related_name="meeting_sponsors")
-    about = models.TextField("About this sponsorship", blank=True, null=True)
+    sponsor = models.ForeignKey("sponsors.Sponsor", on_delete=models.PROTECT)
+    meeting = models.ForeignKey("meetings.Meeting", related_name="meeting_sponsors",
+            on_delete=models.CASCADE)
+    about = models.TextField(u"About this sponsorship", blank=True, null=True)
     about_short = models.CharField(
         "Brief description of sponsorship", max_length=128, blank=True, null=True
```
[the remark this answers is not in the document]
sophm256: After thinking about it for awhile, wouldn't the 'u' be unnecessary in the 'about' field since Python 3.0, the 'str' is by default represented as unicode characters. 

https://docs.python.org/3/howto/unicode.html#the-string-type
From that link:
> The String Type
> Since Python 3.0, the language’s str type contains Unicode characters, meaning any string created using "unicode rocks!", 'unicode rocks!', or the triple-quoted string syntax is stored as Unicode.

chipy_org/apps/sponsors/models.py
sophm256: Wouldn't the 'u' be unnecesary in the 'about' field because, since Python 3.0, 'str' are represented as unicode already? I explain more about it in a previous comment. We are using python 3.6.9 .

chipy_org/apps/sponsors/models.py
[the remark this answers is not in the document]
elmq0022: That's a good catch.  Definitely not needed now that we are firmly in the world of python3.  We can remove those as we come across them.

chipy_org/apps/sponsors/models.py
elmq0022: Actually we're upgrading to 3.8.2 😄 
And yes we can start removing those.

sophm256:
After looking around in the code, the 'sync meetup' button does exist. For it to show up the 'meetup_id' field has to be filled out in the 'meeting' object. So it's a feature, not a bug that the button is not showing up initially. It will show up under certain circumstances.

PR #278, opened 2020-04-19
Redirect fix for login, fix for access to job_post object in admin, and other changes

sophm256:
- Redirect fix for login. When logging into the job posting page, the user will be correctly redirected back to that page after logging in with Social Auth. For the general chipy.org site, the user will be redirected to the 'next' parameter after login (i.e. the page they started from before they had to login)

- Because of an extra parameter (an underscore) in a function in 'admin.py', this prevented an admin from accessing the job post object. This pull request removes that underscore paramater.
 
- Refactored the view function for the job post form to not depend on a hidden form field called 'contact'. Doing it this way make it less error prone. 

- Other small changes

### DIFF
--- a/chipy_org/apps/job_board/views.py
+++ b/chipy_org/apps/job_board/views.py
@@ -11,12 +11,14 @@ from django.views.generic import ListView, DetailView
 from chipy_org.apps.job_board.forms import JobPostForm, JobUserForm, JobProfileForm
 from .models import JobPost
 
+
 @login_required
 def create_job_post(request):
 
     if request.method == "POST":
 
-        job_post_form = JobPostForm(request.POST)
+        job_post = JobPost(contact=request.user)
+        job_post_form = JobPostForm(request.POST, instance=job_post)
         job_user_form = JobUserForm(request.POST, instance=request.user)
         job_profile_form = JobProfileForm(request.POST, instance=request.user.profile)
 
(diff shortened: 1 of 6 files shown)


chipy_org/apps/job_board/views.py
```diff
@@ -11,12 +11,14 @@
 from chipy_org.apps.job_board.forms import JobPostForm, JobUserForm, JobProfileForm
 from .models import JobPost
 
+
 @login_required
 def create_job_post(request):
 
     if request.method == "POST":
```
elmq0022, line 19: Just wondering if you looked at this documentation [long_required](https://docs.djangoproject.com/en/2.2/topics/auth/default/#the-login-required-decorator).  

I see you were able to get this to work with the updates to chipy_org/templates/_login_options.html.  Does this work with python-social-auth?   Maybe a more obvious way as it keep the redirect closer to the view that requires the login.  No need to update.

sophm256: Yes, I did look at that document. I'm not sure what the alternative you are suggesting. **Do you think you can tell me in code?**

Our settings.py defines a `LOGIN_REDIRECT_URL = "/" `and `LOGIN_URL = "/login/"`

 When you try to visit the `chipy.org/job/create/` , it redirects you to login at  `www.chipy.org/login/?next=/job/create/` . This is okay so far. But then it deals with Social Auth.  The `_login_options.html ` somehow tells the browser to go to the home page, instead of where we want to go, which is `/job/create/ `.  After logging in, Social Auth redirects to `chipy.org/#` if you're logging in with Google.

So Social Auth is defining the ultimate destination of where the user ends up.  Please read the first section "Return the user to the original page" in this link
https://python-social-auth.readthedocs.io/en/latest/use_cases.html
This document uses `?next={{ request.path }}` but that would take one to the `'www.chipy.org' `site since that is **path**.  The answer that worked for me is `?next={{ request.GET.next }}`.

elmq0022: Sorry, I misread the document I referenced above and erroneously pointed it out. 

I thought we could just pass the redirect url as an argument to the login_required decorator but that only customize the "next" field name in the context passed to the template.  

Thanks for pointing out the social auth doc.  I wonder why {request.path} didn't work as documented, but I have no issues as I tested it and it does work as intended.    

Kudos on the job board!

PR #279, opened 2020-04-29
updates config.yml to python 3.8

elmq0022:
Looks like I broke the CircleCI build in the Django 2.2 / Python 3.8 upgrade.  Hopefully the updates here fix it.


elmq0022:
Thanks for looking at it.  We won't know for sure until we merge to master
and get CircleCI to run.  I don't anticipate any problems.

PR #281, opened 2020-04-29
fixes about link on homepage

elmq0022:
Changes `name=about` to `id=about` so the about link on the homepage works.


elmq0022:
I updated the rest of `name=` for consistency on the homepage.

PR #283, opened 2020-05-01
Don't 500 on blank form submissions to RSVP

### LINKED ISSUE #282: Login without field throws and Error
If the user does not fill in the fields below. An error is thrown when they click RSVP.  The user should be prompted for the information instead being shown and error page.  

login

error

Maybe we can add some JavaScript validation around the fields?

raymondberg:
## Problem
Fixes #282 

We're getting 500 errors on form submissions.

## Solution

1. Verify we can test a working form
2. Write a test of a malformed form
3. Verify that 500 is occurring
4. Add required fields

New behavior: 

Screen Shot 2020-04-30 at 7 17 14 PM

Additionally, if someone bypasses the JS subscriptions it will redirect to a (very ugly) page showing form submission errors.

### DIFF
--- a/chipy_org/apps/meetings/tests.py
+++ b/chipy_org/apps/meetings/tests.py
@@ -180,3 +180,25 @@ class MeetingTitleTest(TestCase):
             when=datetime.date.today(), custom_title="Main Custom Title"
         )
         self.assertEqual(meeting.title, "Main Custom Title")
+
+@override_settings(STATICFILES_STORAGE=global_settings.STATICFILES_STORAGE)
+def test_rsvp_works_for_anonymous_user(client):
+    meeting = Meeting.objects.create(
+        when=datetime.date.today() + datetime.timedelta(days=1)
+    )
+    route = reverse("rsvp")
+    response = client.post(route, data={
+        "meeting": meeting.id,
+        "first_name": "Some",
+        "last_name": "Body",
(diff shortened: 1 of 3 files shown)


chipy_org/settings_test.py
```diff
@@ -3,7 +3,7 @@
 
 DATABASES = {"default": {"ENGINE": "django.db.backends.sqlite3", "NAME": ":memory:", "TEST": {}}}
 DEBUG = True
-ADMINS = "admin@chipy.org"
+ADMINS = ["admin@chipy.org"]
```
raymondberg, line 6: Admins should always be plural, setting it with a string caused 10s of layers of failures in the test stack trace when I was trying to debug.

chipy_org/apps/meetings/tests.py
```diff
@@ -180,3 +180,25 @@ def test_main_meeting_with_custom_field(self):
             when=datetime.date.today(), custom_title="Main Custom Title"
         )
         self.assertEqual(meeting.title, "Main Custom Title")
+
+@override_settings(STATICFILES_STORAGE=global_settings.STATICFILES_STORAGE)
+def test_rsvp_works_for_anonymous_user(client):
```
JoeJasinski, line 185: Nice adding some tests

PR #284, opened 2020-05-03
adds -j 0 to pylint command

elmq0022:
The -j 0 automatically detects the number of cores and runs pylint in parallel.  On my system using the -j 0 option reduced lint time from around 30 seconds to 22 seconds.


elmq0022:
You're right.  This may not make an impact for everyone as the speed is
solely based on the availability hardware.

PR #285, opened 2020-05-03
Add isort to Format Command and Format Recent Updates

elmq0022:
Applying black formats missed in the last two deploys. Thanks!

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,6 @@
+[settings]
+line_length=99
+skip=manage.py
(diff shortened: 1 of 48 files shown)


elmq0022:
I have some scope creep here. I've gone ahead and added isort to the format command.  isort and black will fight a bit about trailing commas on imports, so I run Black last to keep the "minimal diffs" a consistent requirement.

elmq0022:
That's a great update. I'll added it before we merge this request.

Issue #287, opened 2020-05-04
Allow Speakers to Submit Updates to their talk submission for approval


JoeJasinski:
As a Speaker, I would like to be able to update my talk proposal so that a board member can review and approve my changes. 

I'm working on a PR that would allow users to view their talk submissions and then submit update drafts to them. This will allow speakers and admins to review the diff of changes. Admins can then choose to approve the changes, which would sync the changes to the live.

PR #288, opened 2020-05-04
adding meeting topic draft feature

JoeJasinski:
This is a PR to address the following. More info to come, but creating a placeholder PR.

See the attached issue for information on how this works.
https://github.com/chicagopython/chipy.org/issues/287

This PR requires one migration to add a new TopicDrafts table.


sophm256:
@JoeJasinski Really nice job on this feature! I left you a comment on Slack that talks a little bit more about this feature. Thanks for all of your hard work and your attention to detail :smile:

chipy_org/apps/meetings/models.py
```diff
@@ -229,7 +233,80 @@ def __str__(self):
...
+            if getattr(self, fld) != getattr(other, fld):
+                return False
+        if other.id != self.topic.id:
+            return False
+        return True
+
+    def __hash__(self):
+        if self.pk is None:
+            raise TypeError("Model instances without primary key value are unhashable")
+        return hash(self.pk)
```
elmq0022, line 309: I think this is causing an error as it's the only hash function implemented / changed here.  I am unable to delete topics through the admin.  

image

followed by:

image

elmq0022: Also, I did confirm this functionality does work on master.

chipy_org/apps/meetings/email.py
```diff
@@ -37,3 +37,19 @@ def send_meeting_topic_submitted_email(topic, recipients):  # pylint: disable=in
...
+        context = {"topic": draft.topic, "draft": draft, "site": Site.objects.get_current()}
+        subject = "Chipy: Updates to Meeting Topic Submitted"
+        from_email = "DoNotReply@chipy.org"
+        text_content = plaintext.render(context)
+        html_content = htmly.render(context)
+        msg = EmailMultiAlternatives(subject, text_content, from_email, recipients)
+        msg.attach_alternative(html_content, "text/html")
+        msg.send()
+    except Exception as e:
+        logger.exception(e)
```
elmq0022, line 55: I think we have pretty similar code in three places now, here, the RSVPs, and the contact.  I wonder if it's time extract this as a method in a library.

chipy_org/apps/meetings/models.py
elmq0022: Also isn't this the implementation the default?  does it need to be overridden specifically?  

[django model hash](https://docs.djangoproject.com/en/2.2/ref/models/instances/#hash)

elmq0022: [code](https://docs.djangoproject.com/en/2.2/_modules/django/db/models/base/#Model.__hash__)

elmq0022:
Also I'm unable to save an edit to a main meeting

image

image

Not sure how this is related.

JoeJasinski:
@elmq0022 thanks for the feedback. I originally added that `__hash__` method to try to resolve this delete issue, and thought that it had. However, looks like it needs another look.  You are right, it is the default implementation; it had looked like the default implementation wasn't being used due to some of the inheritance, so I set it explicitly there. I'll take another look at it this weekend. 

I agree that the email logic could be consolidated. I'll create another story where we can consider how we want to make some standard email logic.

elmq0022:
Sounds great.  I'm looking forward to the updated version 👍.

elmq0022:
@JoeJasinski, I missed your update.  Do you need me to take a look at this one?

PR #290, opened 2020-05-13
Updated README and Makefile

sophm256:
In the Makefile, I added make instructions to create a superuser and view logs. Also added seeing a preview of what the isort function would look like under the 'format-check'.

In the README, I did the following changes:
- Told the user to fork the repo, and then clone it.
- Removed the info about having errors after running the 'make migrate' command, because we no longer have errors (Thanks to @elmq0022 for ironing out those errors)
- Added info regarding 'make format' and 'make format-check'
- Added info regarding 'make shell' and 'make log'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,8 +57,8 @@ format-check:
 
 setup: setup_env build
 
-super: 
+superuser: 
 	docker-compose exec web ./manage.py createsuperuser
 
-log: 
+tail-logs: 
 	docker-compose logs -f web
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,8 @@ format-check:
 
 setup: setup_env build
 
-super: docker-compose exec web ./manage.py createsuperuser
+super: 
+	docker-compose exec web ./manage.py createsuperuser
 
-log: docker-compose logs -f web
+log: 
+	docker-compose logs -f web
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,11 @@ format:
 	docker-compose exec web black .
 
 format-check:
+	docker-compose exec web isort -rc -tc --atomic --diff .
 	docker-compose exec web black --diff .
 
 setup: setup_env build
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -52,6 +52,13 @@ format:
...
 format-check:
+	docker-compose exec web isort -rc -tc --atomic --diff .
 	docker-compose exec web black --diff .
 
 setup: setup_env build
+
+super: 
+	docker-compose exec web ./manage.py createsuperuser
+
+log:
```
raymondberg, line 63: I think I prefer `tail-logs` here

Makefile
```diff
@@ -52,6 +52,13 @@ format:
 	docker-compose exec web black .
 
 format-check:
+	docker-compose exec web isort -rc -tc --atomic --diff .
 	docker-compose exec web black --diff .
 
 setup: setup_env build
+
+super:
```
raymondberg, line 60: Can we `superuser` this instead? Super is ambiguous.

Makefile
```diff
@@ -52,6 +52,13 @@ format:
 	docker-compose exec web black .
 
 format-check:
+	docker-compose exec web isort -rc -tc --atomic --diff .
```
raymondberg, line 55: This is a new one for me, thanks!

Makefile
sophm256: Sure. Will do.

Makefile
sophm256: Ok.

Makefile
sophm256: @elmq0022 has a current PR https://github.com/chicagopython/chipy.org/pull/285
that adds `docker-compose exec web isort -rc -tc --atomic .` to the Makefile 'format'. So I decided to add the accompanying diff command so people can preview.

sophm256:
Thanks for reviewing my PR @raymondberg ! I appreciate your comments. 😄

PR #299, opened 2020-06-29
changes field length_minutes to length

elmq0022:
This fixes the error sent to the chipy organizers this morning around 2:35 AM.  There are inconsistent names for the length of a presentation.


elmq0022:
I don't know if anyone relies on the format of the api.  We can change everything back to length instead of length_minutes.  There's  a few more fields to update and a migration would be required.  Might be better to do this as I don't know who is using the api.

elmq0022:
okay - why isn't git picking up migrations any more?

elmq0022:
We should also add a test for the API views.

Issue #270, opened 2020-03-29
Job Board MVP


alysivji:
This ticket represents the feature -- will need to write stories and connect them to this epic.

----

### Background

ChiPy has a [Jobs referral](https://www.chipy.org/pages/referrals/) program. There is also a #jobs channel in our Slack workspace.

### Proposed Feature

A job board with open positions that we can host on ChiPy.org would be a great compliment to our jobs program. It would benefit to our community in many ways:
- help members find new opportunities
- help sponsors find new employees
- fund activities and events to futher ChiPy's mission

### Minimum Requirements

- have a page showing a list of jobs -- high level overview
- can click into each job to get additional details about each job
- anybody can submit a job, only authorized users can approve jobs to appear
  - submitters have to accept the guidelines of ChiPy's [Jobs referral](https://www.chipy.org/pages/referrals/) policy
- important fields: `company`, `title`, `description` (accept markdown), `contact` (`name`, `email`, `title`), `link` (to company page)
  - definitely missed somethings here
- Are you an external recruiter? checkbox
- Is your organization interested in hosting a meetup? checkbox
- `created_at` and `updated_at` columns in all tables (if we don't already do that)

### Nice to haves

- "featured jobs" -- have a way for sponsors job
- search; good feature to add once we have more jobs 
- periodically post jobs from website to Twitter, LinkedIn, or Slack
  - it would be great if this jobs board is the only way people can promote jobs on our Slack

### Open Questions

- what are the different states a job posting can have?
- how long should each job posting be shown for?
- should we limit the number of job posting from a single organization?
- should we allow external recruiters to post jobs?

alysivji:
@eevelweezel do you have any thoughts on this?

eevelweezel:
I think we maybe should have posts auto-expire (say, 60 days?), and give submitters some mechanism to request removal of their post.  For sponsors, maybe we could provide a way to override this, and allow them to keep their post up for longer (maybe this ties into "featured jobs")? As for external recruiters, I'm fine with allowing them, provided the they agree to follow our guidelines AND provide the name of the company.

I'd say let's keep it to one active post per position.  While I like the idea of locking down the Slack channel to only posts from the jobs board, that wouldn't necessarily stop other forms of abuse, such as staffing firms outside of Chicago contacting employers about remote workers.  If we did lock down Slack, would members still be able to post their resumes?

elmq0022:
I agree with the auto expire feature.  We could definitely make this
[...]

PR #306, opened 2020-07-29
Implement min height in job board (#302)

sophm256:
## Problem
@alysivji pointed out that the footer was crowding the top. 

https://github.com/chicagopython/chipy.org/issues/270#issuecomment-624396508

## Solution
This PR gives a minimum height for all the template pages in the Job Board, which pushes the footer down. In the very far future, there'll be a site redesign for chipy, which will give us a more long term solution.
footer


alysivji:
🎉

PR #316, opened 2020-08-24
Job board bug - Only set job post approval and expiration on initial approval

cdfuller:
While browsing the code, I noticed that a JobPost will have its approval data and expiration date updated whenever it's saved. I assume this isn't intended functionality since it isn't explicit.

### DIFF
--- a/chipy_org/apps/job_board/models.py
+++ b/chipy_org/apps/job_board/models.py
@@ -105,7 +105,8 @@ class JobPost(CommonModel):
             self.status_change_date = datetime.datetime.now()
             self.__original_status = self.status
 
-        if self.status == "AP":  # if post is approved, set the approval date and expiration date
+        # if post is approved, set the approval date and expiration date
+        if self.status == "AP" and not self.approval_date:
             self.approval_date = datetime.datetime.now()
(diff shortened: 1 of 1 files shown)


elmq0022:
I can merge and deploy provide the selected reviewers have provided feedback.

sophm256:
There is one condition that hasn't been yet accounted for. If the post has been approved AND the expiration date has been set AND the days_to_expire is change, then the expiration_date has to be recalculated. (This is because the expiration_date is dependent on the days_to_expire and approval_date.)

**Please change your code according to my gist to account for this**.  The gist is: 
https://gist.github.com/sophm256/3a4f5ee598490fd39d2b6fc58f16c017

The easiest thing to do is just to copy and replace the whole gist to replace the __init__() and save(). (Please do not change the formatting because this code has been run through lint and formatted with Black.)

sophm256:
I have edited my gist to account for another condition.

sophm256:
@alysivji Sorry, I accidentally re-requested a review from you.

sophm256:
@cdfuller , please disregard my earlier messages regarding having to change your PR. I'm going to try to figure out a way to keep our code separate so that we can get credit separately. So you don't have to copy and paste from my gist. Sorry for the confusion!

PR #318, opened 2020-08-25
Fix ability to set days_to_expire and expiration_date

sophm256:
This PR address the following issue:
- Adds the ability to properly save the 'days_to_expire' field. This way if the admin changes the 'days_to_expire' field to shorten or lengthen the time for the post to show up, the 'expiration_date' will properly be recalculated.


chipy_org/apps/job_board/models.py
```diff
@@ -105,11 +106,32 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+        # then recalculate expiration_date.
+        elif (
+            self.status == "AP"  # pylint: disable=bad-continuation
+            and self.approval_date  # pylint: disable=bad-continuation
+            and self.__original_days_to_expire  # pylint: disable=bad-continuation
+            != self.days_to_expire  # pylint: disable=bad-continuation
+        ):
+
+            self.expiration_date = self.approval_date + datetime.timedelta(days=self.days_to_expire)
+            self.__original_days_to_expire = self.days_to_expire
```
raymondberg, line 133: I'm having a hard time following this logic. Is it easier to write a nested if block here?

```python
if self.status == "AP":
  ## update expiration date
  if not self.approval_date: #set approval date
  if self.__original_days_to_expire != self.days_to_expire: #set odte
```
     ...

sophm256: @raymondberg Thanks for the comment. Your code isn't equivalent to my code. 
One of the conditions in my code is `if self.approval_date` whereas yours tests for `if NOT self.approval_date`. 
......................................................................................
IN THE BEGINNING 

Here is a hypothetical scenario. After a user creates a post, the status will be `submitted`. When the admin first changes the status to `approved`, the 'if' statement on line 118 will go into effect and the approval_date and expiration_date are set.

```
if self.status == "AP" and not self.approval_date:
    self.approval_date = datetime.datetime.now()
    self.expiration_date = self.approval_date + datetime.timedelta(days=self.days_to_expire)
```
............................................................................................................................
LATER ON

Let's say, later on, the admin wants to extend the job post so that it will show up for more than 60 days. Let's say they want 90 days. The admin will then change the `days_to_expire` field in the admin panel. After the admin presses **save**, the condition for the 'elif' statement on line 132 will now be satisfied.

At this point, the post is in an `approved` state. The approval_date already exists at this point, because we set it the very first time when the post was approved. Since the `days_to_expire` is a new value of **90** and `__original_days_to_expire` is **60**, then they are not equivalent.

(This is the code without the automatic formatting and lint disables.)
```
# If post is approved AND expiration_date has been set AND days_to_expire is changed,
# then recalculate expiration_date.

elif ( self.status == "AP"  and self.approval_date and self.__original_days_to_expire  != self.days_to_expire ):
    self.expiration_date = self.approval_date + datetime.timedelta(days=self.days_to_expire)
    self.__original_days_to_expire = self.days_to_expire
```
We have to recalculate the expiration_date because it depends on the fields 'approval_date' (which is the same value as it was before) and 'days_to_expire' (which value has now changed).  

`__original_days_to_expire` is set to  `self.days_to_expire` to keep track of if `days_to_expire` will change in the future.

I think the code in the actual PR looks confusing because of the automatic formatting that that had to be done and all of the lint disables! :smile:

sophm256: @raymondberg, For readibility's sake, I can turned it into a nested statement. Thanks for your great feedback! :apple:
```
elif self.status == "AP" and self.approval_date:
     
    if self.__original_days_to_expire != self.days_to_expire:
        
        self.expiration_date = self.approval_date + datetime.timedelta(
            days=self.days_to_expire
            )
        
        self.__original_days_to_expire = self.days_to_expire
```

raymondberg: Interesting. I think we might need some real good model unit tests here to document what is expected and call out this bug. I'm not sure I understand entirely (which means tests would be great to enforce what you wrote up). But I am curious if something like this would work: https://github.com/chicagopython/chipy.org/compare/master...raymondberg:further-expiration-date-rb?expand=1

sophm256: One reason why my code is complicated is that the expiration_date and days_to_expire field need to be in sync with each other.  I wanted the admin to be able to adjust the number of days as an integer, because it's easy to think in whole numbers about how many days a post should be shown. The expiration_date is then calculated programmatically since humans can't easily calculate dates accurately.  I wanted the computer to be able to use the expiration_date as a filter in the SQL query, where post are shown` if today's date <= expiration date `. I've thought about making `expiration_date` into a property, but you can't filter a Django queryset off of a property.

**I think your code is a great revision.** 

Here is how I would have written it: 
https://gist.github.com/sophm256/93f437a151b632ae47f6d20acbce5fd9

**That being said, I think we should go with your version of the code**. The only difference in your code and mine is that an extra "update_expiration_date()" is run an extra time in your code in the following scenario:

```
Admin sets the status to 'approved' for the first time.
Admin changes the 'days_to_expire'.
Admin then presses 'save' button. 
```

THE DIFFERENCE BETWEEN MY CODE AND YOUR CODE
In my code in that scenario, the update_expiration_date() would only run once overall.  The expiration date is calculated with the correct values the first time under the if statement `if not self.approval_date:` line 119 from my gist. 

In your code, the update_expiration_date runs twice: the first time when it hits `if not self.approval_date:` (line 119) and a second time when it hits `if self.__original_days_to_expire != self.days_to_expire:` (line 127 from your code) . Technically, `update_expiration_date()` doesn't need to run twice, because it was calculated the first time correctly in this particular scenario.

........................................................

I've made a new commit with essentially what you've suggested. **I also have added some more code to account for another condition (line 131 to 136 in my 3rd commit)**. If the admin approves the post but then changes the status to something different, it changes the approval_date and expiration_date back to None. **I will try to write up some tests eventually for this in the future in a separate PR.** Thanks again for your help!

elmq0022: Okay I am late to the party, but could you just not calculate an expire date and avoid syncing it.  Just keep the approval date and the number of days to expire value.   I would push the work to view that only displays active posts, i.e approved and approval date + days to expire > datetime.now.  

May make things cleaner.

sophm256:
Thanks @elmq0022 for your suggestion!  In a new commit, as suggested, I took out the expiration_date field. The model code got simpler and the queryset code in the view got a little more complicated. (I think I avoided doing this because I wasn't sure how to get the query right. But after some experimenting, I got it to work!) I also took out the `status_change_date` field and the related code in init() and save() because we weren't using it. :cake:

@raymondberg Let me know if you still think model tests are needed. But the model has been simplified and the confusing code is probably gone 🍦

chipy_org/apps/job_board/models.py
```diff
@@ -100,22 +94,19 @@ class JobPost(CommonModel):
...
+                self.approval_date = datetime.datetime.now()
 
-        # if post is approved, set the approval date and expiration date
-        if self.status == "AP" and not self.approval_date:
-            self.approval_date = datetime.datetime.now()
-            self.expiration_date = self.approval_date + datetime.timedelta(days=self.days_to_expire)
+        # If post was approved but then changed to a different status,
+        # set the approval_date back to None.
+        elif self.status == "RE" or self.status == "SU":
+            if self.approval_date:
```
raymondberg, line 108:
```suggestion
        elif self.approval_date:
```

chipy_org/apps/job_board/views.py
```diff
@@ -171,15 +171,26 @@ class JobPostList(ListView):
...
+            & Q(is_sponsor=True)
+            & Q(approval_date__gte=now - (datetime.timedelta(days=1) * F("days_to_expire")))
+        ).order_by("-approval_date")
         other_job_posts = JobPost.objects.filter(
-            Q(status="AP") & Q(is_sponsor=False) & Q(expiration_date__gte=datetime.datetime.now())
-        ).order_by("-id")
+            Q(status="AP")
+            & Q(is_sponsor=False)
+            & Q(approval_date__gte=now - (datetime.timedelta(days=1) * F("days_to_expire")))
+        ).order_by("-approval_date")
```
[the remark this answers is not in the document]
sophm256: What does that look like? I'm not familiar with having queries in the model.

chipy_org/apps/job_board/models.py
sophm256: Good suggestion. Will do.

chipy_org/apps/job_board/views.py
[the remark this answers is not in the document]
raymondberg: It could be a classmethod, something like `JobPost.sponsored_posts()`.

elmq0022:
I'll second a class method for this use case.  Nicely bundles all the
available queries with the related class.

chipy_org/apps/job_board/views.py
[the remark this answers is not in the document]
sophm256: ok team, I will move this into a class method :thumbsup:

chipy_org/apps/job_board/models.py
```diff
@@ -130,3 +122,47 @@ def days_elapsed(self):
             return days_elapsed_from_posting
```
elmq0022, line 122: You can remove line 130 and just return delta.days.  Also would you want to return a timedelta object?  You seem to be casting the in to a timedelta most places.  
```python
return timedelta(days=delta.days)
```

chipy_org/apps/job_board/models.py
```diff
@@ -130,3 +122,47 @@ def days_elapsed(self):
...
+
+        other_job_posts = JobPost.objects.filter(
+            Q(status="AP")
+            & Q(is_sponsor=False)
+            & Q(approval_date__gte=now - (datetime.timedelta(days=1) * F("days_to_expire")))
+        ).order_by("-approval_date")
+
+        # I put the two groups of job posts back together so they can processed
+        # by the same loop in the template.
+        job_posts = list(chain(sponsored_job_posts, other_job_posts))
```
elmq0022, line 166: Do you need to break up the queries? Could you use group_by to group the sponsored and other job posts.  

Another way may be three method.  The first method would be `approved` to get all the approved posts.   Then you use  two more methods which are `approved_and_sponsored` and `approved_and_not_sponsored' which rely on the code from `approved`.   Then you can easily make the calls below in your code.    

1. JobPost.approved.all()
2. JobPost.approved_and_sponsored.all()
3. JobPost.approved_and_not_sponsored.all()

**Side Note:**
@raymondberg  and I both asked you to use a classmethod for this.  However, I recently found out about  [Django's Manager Class](https://docs.djangoproject.com/en/3.1/topics/db/managers/).  I won't ask you to refactor using the Django manager.  I do want to point out what seems to be the "Django way" of doing this task.

chipy_org/apps/job_board/models.py
```diff
@@ -130,3 +122,47 @@ def days_elapsed(self):
...
+            expiration_date = self.approval_date + datetime.timedelta(days=self.days_to_expire)
+            return expiration_date
+        else:
+            return None
+
+    @classmethod
+    def approved_posts(cls):
+
+        # I've split these into two queries in anticipating that there might be
+        # different ordering or filtering based on sponsored vs non-sponsored
```
elmq0022, line 142: maybe this should read approved_and_active to denote that they are in an approved state, but they also have not expired.

elmq0022 (approved):
I think this is a good fix.  I did leave some comments that you could address, but I don't think they are worth holding up a deploy of this fix.  

We may want to do some refactoring later though.  Thanks.  :beers:

PR #320, opened 2020-09-04
Change path parameters to query string parameters

sophm256:
This PR refactors code so that instead of using path parameters (such as in `after-submit/<int:action>/` where 'action' is the path parameter), the "after-submit-job-post" page uses query string parameters `(such as http://chipy.org/job-board/after-submit/?action=create)` to control code logic.


raymondberg (approved):
Looks reasonable to me.

PR #322, opened 2020-11-16
Changed homepage template to say main meetings are occurring online

sophm256:


elmq0022 (approved):
I think this is fine for the time being.  But it would be better to manage this content with something other than a deploy.  

I know we will start up actual meetings sometime in the future and it would be nice to give administrators the ability to update this through the admin.  Could we use a flat page here?

raymondberg (approved):

PR #323, opened 2020-11-22
removes general sponsor in favor of sponsor

elmq0022:
We don't really make a distinction among sponsor, general sponsor, and meeting sponsor.  So, I propose we delete the general sponsor model.  

Also, I think we could or should have a Meeting Host instead of Meeting Sponsor.  That object would be more descriptive and in line with the intent of the object.  A business or location could host a Chipy meeting, but they might not sponsor Chipy.  This gets a little fuzzy through, so I've left Meeting Sponsor allow for now.


chipy_org/apps/main/views.py
```diff
@@ -37,9 +37,7 @@ def get_context_data(self, **kwargs):
         context.update(kwargs)
 
         context["other_meetings"] = self.get_non_main_meetings(num=3)
-        context["general_sponsors"] = GeneralSponsor.objects.all().order_by(
-            "sponsor__sponsor_group__list_priority", "sponsor__name"
-        )
+        context["sponsor_groups"] = SponsorGroup.objects.prefetch_related("sponsors")
```
sophm256, line 40: Nice work overall! 

We sort of want something analogous to the old lines 40 to 41 that you deleted. That is, we want ordering by 'list_priority' from the `SponsorGroup` and then ordering by 'name' of the `Sponsor`.
............................................................................................................

1)Please change new line 40 to become:
`context["sponsor_groups"] = SponsorGroup.objects.prefetch_related("sponsors").order_by("list_priority")`

2)Then, `make format` and` make lint` since that line will be too long.

After ordering by 'list_priority' of the` SponsorGroup`, you will NOT need to change anything else to order by the 'name' of the `Sponsor`.  This is because that's already handled by the 'meta' class in the `Sponsor` model. (See chipy_org/apps/sponsors/models.py , lines 59-60 )

sophm256 (changes requested):

chipy_org/apps/main/views.py
elmq0022: Just added the list priority to the sponsor groups as well to resolve the issue.  Also cleaned up tests a little like we did for the job board.  Good catch.  Thanks!

chipy_org/apps/sponsors/tests.py
```diff
@@ -1,22 +1,23 @@
 # pylint: disable=invalid-name, duplicate-code, redefined-outer-name
 import pytest
 from django.conf import global_settings
-from django.test import Client, override_settings
```
sophm256, line 4: Why was 'Client' in this import statement removed as well as in line 18, 19? It seems that the test functions are dependent on the client because they're passed in as a parameter.

Also, I've run the tests and they seem to be passing instead throwing errors so I'm puzzled. Is there something I'm not seeing?

elmq0022:
The client fixture is provided by pytest.  So I was duplicating code.

Looks like circle ci passed all the tests.

You could try locally as well or compare to the job board tests.

Thanks 😊

sophm256 (approved):

raymondberg (approved):
Excellent Work. 🍔

elmq0022:
Thanks!

PR #326, opened 2020-12-05
uses custom model manager instead of classmethod

elmq0022:
This refactors the on-the-fly fixes from our last deploy.

- Uses custom model.Manger
- Uses queryset vs get_queryset in ListView
- Uses ordering in ListView
- Deletes IntegerField days_to_expire in favor of DurationField time_to_expire (migration to apply for associated deploy)


sophm256 (approved):
@elmq0022 Great job on refactoring! Everything works and is in order. 🍦

chipy_org/apps/job_board/models.py
```diff
@@ -33,6 +32,15 @@
 ]
 
 
+class ApprovedAndActive(models.Manager):
+    def get_queryset(self):
+        return (
+            super()
+            .get_queryset()
+            .filter(Q(status="AP") & Q(approval_date__gte=Now() - F("time_to_expire")))
+        )
```
sophm256, line 41: @elmq0022 I would prefer that the ordering be here (by chaining two `order_by`) rather than in the Listview (line 186 in views.py) to make it clear what's going on in the query.

So line 41 would read `.order_by("-is_sponsor").order_by("-approval_date")` 
I'm not sure if this line is too long for our linter so you might need to put them on two lines.

chipy_org/apps/job_board/views.py
```diff
@@ -180,14 +180,12 @@ def get_queryset(self):
 
 
 class JobPostList(ListView):
-
     model = JobPost
     context_object_name = "job_posts"
     template_name = "job_board/job_post_list.html"
+    ordering = ("-is_sponsor", "-approval_date")
```
sophm256, line 186: Please remove line 186 here if you put the ordering for the queryset in the 'ApprovedAndActive' function in models.py

chipy_org/apps/job_board/models.py
elmq0022: My thought process was to create a model manager that specifically queries the active (i.e. not expired) and approved job posts while delaying an expensive order/sort operation until needed.  

For example if we want a count of all approved / active job posts (JobPost.approved_and_active.count()), we don't need to call sort.  Also what if we want a different order for the sort or filter to just sponsored jobs?  

Also, it's easy to off load the sort to the class based view.

sophm256: Ok, I see what you're saying. In that case, you don't need to modify it. 👍

nickymarino (approved):
Nice! Looks good to me 🎈

PR #333, opened 2020-12-24
Adds verbose names to the RSVP model.

elmq0022:
I got sick of seeing Rsvp and Rsvps in the admin.  Should be an easy review.  :sweat_smile:

### DIFF
--- a/chipy_org/apps/meetings/models.py
+++ b/chipy_org/apps/meetings/models.py
@@ -257,6 +257,8 @@ class RSVP(CommonModel):
 
     class Meta:
         ordering = ["-meeting", "last_name", "first_name"]
+        verbose_name = "RSVP"
+        verbose_name_plural = "RSPVs"
 
     def clean(self):
         if not self.user and not self.email:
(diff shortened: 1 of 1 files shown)


chipy_org/apps/meetings/models.py
```diff
@@ -257,6 +257,8 @@ class RSVP(CommonModel):
 
     class Meta:
         ordering = ["-meeting", "last_name", "first_name"]
+        verbose_name = "RSVP"
+        verbose_name_plural = "RSPVs"
```
sophm256, line 261: Please change line 251 to `RSVPs`. Thanks!

sophm256 (changes requested):
Plea

chipy_org/apps/meetings/models.py
sophm256: Actually, you can take out line 251 because by default, Django adds an 's' for the plural form of verbose name if none is specified.

elmq0022: Dang, typing is hard.  I've made the update.  Thanks!

sophm256 (approved):
Great! Easy peasy review. :cake:

raymondberg (approved):
🍔

PR #338, opened 2021-01-06
Rip Out Old JS/CSS/Less resources

raymondberg:
## Problem

@sophm256 is going to work on a redesign, but it's a little confusing with all 
our ancient .less and .js files laying everywhere. They aren't even used. We
think we're going to go with a CDN model for the JS libs, but at least we should limit
ourselves to just the JS that we need to do page renders.

## Solution

We've isolated all the JS/CSS that is rendered in our views (not the less version, 
which is not referenced anywhere). And the set of JS/CSS we actually use is tiny.

Burn everything else (trust me, it still looks pretty good)

image

@elmq0022 @nickymarino


elmq0022 (approved):
I pulled it down and deployed locally.  Seems fine to me.  Styles are a little different, but let's move forward with this.

Issue #341, opened 2021-01-15
Changing Anonymous RSVP after Meeting Closes Raises Error


elmq0022:
Anonymous RSVPs can be changed via a secret alpha numeric key of 40 characters.  If a user tries to change an anonymous RSVP after the meeting is closed by navigating to /meetings/rspv/[a-z0-9] an error is raised instead.    

At minimum the site should instead at least show a 404 page instead of raising an error.  

I think it would be highly preferably to let the user know that RSVPs for that meeting have been closed and can no longer be updated.

elmq0022:
in UpdateRSVP view
    
```python
def get_object(self, queryset=None):
    obj = get_object_or_404(RSVPModel, key=self.kwargs["rsvp_key"])
    if not obj.meeting.can_register():
        messages.error(self.request, "Registration for this meeting is closed.")
            return redirect(reverse_lazy("home"))
    return obj
```
The issue is the return redirect line.  this returns a redirect object and not an RSVP object. 

The logic to check it a meeting can still be registered should be moved to the get method and cal an HttpResponseRedirect there.

PR #342, opened 2021-01-15
adds custom get method to UpdateRSVP class

elmq0022:
This will work, but please review and let me know if you have a better approach.  Right now I'm hitting the database twice for the RSVP.  Might be better to just completely rewrite the get method with out using super.


elmq0022:
I added a couple tests around the update rsvp view to make sure it redirects when registration for the meeting is closed.

elmq0022:
Solves #341

raymondberg (approved):
🍔

sophm256 (approved):

PR #344, opened 2021-01-18
Add slim style and implement it on job board

sophm256:
## Problem

We've implemented the redesign theme for the homepage, but it's a special page which won't be 
used everywhere.

## Solution

Implement a slim.html theme which will be used for all non-homepage pages. We tested this out 
by using it on the Job Board's list page with a nifty little hack to generate a preview.

The results are just stunning, no?


raymondberg (approved):
👍  We paired, and it's great!

PR #349, opened 2021-01-22
Change column behavior so that 'featured sponsor' section stacks on top of 'next event' section

sophm256:
-In homepage, change column behavior so that 'featured sponsor' section stacks on top of 'next event' section in tablet and mobile mode

-Fix icon size in footer


raymondberg (approved):
Heck yes, can't wait to try this out!

PR #352, opened 2021-01-30
Fix sticky navbar; Place containers to fix extra large screen views

sophm256:


raymondberg (approved):
These look like great improvements!

PR #368, opened 2021-02-13
Adds django rq

elmq0022:
This pull request will add RQ and the django-rq packages to chipy.org.  I also put the contact email on a background worker as a demo.  

I need to look into how to run the tests with this setup though as it is going to provide for some additional complexity.  Contact form email test currently fails with this setup.  

[my heroku site](https://chipy-ace.herokuapp.com/)

### DIFF
--- a/chipy_org/settings.py
+++ b/chipy_org/settings.py
@@ -355,10 +355,18 @@ LOGGING = {
         "simple": {"format": "%(levelname)s %(message)s"},
     },
     "handlers": {
-        "null": {"level": "DEBUG", "class": "logging.NullHandler",},
+        "null": {
+            "level": "DEBUG",
+            "class": "logging.NullHandler",
+        },
         "console": {"level": "DEBUG", "class": "logging.StreamHandler", "formatter": "verbose"},
     },
-    "loggers": {"chipy_org": {"handlers": ["console"], "level": "INFO",}},
+    "loggers": {
+        "chipy_org": {
+            "handlers": ["console"],
+            "level": "INFO",
+        }
+    },
 }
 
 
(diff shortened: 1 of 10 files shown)


chipy_org/settings.py
```diff
@@ -362,3 +363,11 @@ def env_list(key, defaults=None, delimiter=","):
 
 
 STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
+
+
+# django-rq settings
+RQ_QUEUES = {
+    "default": {"URL": os.getenv("REDISTOGO_URL", "redis://redis:6379/0"), "DEFAULT_TIMEOUT": 500,},
+    "high": {"URL": os.getenv("REDISTOGO_URL", "redis://redis:6379/0"), "DEFAULT_TIMEOUT": 500,},
+    "low": {"URL": os.getenv("REDISTOGO_URL", "redis://redis:6379/0"), "DEFAULT_TIMEOUT": 500,},
```
raymondberg, line 372: This feels like an odd name for this environment variable. Is this documented in the code somewhere as being a default or convention?

raymondberg (approved):
This is really clean, nice work! 🍔

chipy_org/settings.py
elmq0022: Yes, these environment variables are required by the django-rq package.

elmq0022:
I will probably end up canceling this pull request as we would need a second dyno for the site.  This would cost additional $$$.  :man_shrugging:   

Let me know if you have a good use case such that we could justify the additional cost.   Honestly, though we can probably just schedule most of the intensive tasks we want to add with chron (scheduler) to run on off hours.

elmq0022:
This is way too old to merge.

PR #395, opened 2021-03-27
Change get_meeting() so that it gets the next meeting from all of the meetings

### LINKED ISSUE #382: Fix homepage to show next meeting correctly
On the homepage chipy.org, the 'next meeting' blue button does not show the date of correct next meeting. For example, algo sig had a meeting today and it did not show that. Fix it so that date of the correct next meeting shows. 

Also, fix it so that the correct next meeting shows up in the 'next meeting' section.

sophm256:
Fixes #382

-Change get_meeting() so that it gets the next meeting from all of the meetings, not just the next __main__ meeting

-Before this fix, the homepage showed the next meeting to be the an upcoming ChiPy__main__ meeting. After this fix, it shows the next meeting to be a meeting from the entire pool of possible meetings.

See also line 67 from https://github.com/chicagopython/chipy.org/blob/main/chipy_org/apps/meetings/views.py to see where the 'next_meeting' is added to the context.

### DIFF
--- a/chipy_org/apps/main/views.py
+++ b/chipy_org/apps/main/views.py
@@ -17,16 +17,9 @@ from chipy_org.apps.sponsors.models import Sponsor, SponsorGroup
 class Home(TemplateView, InitialRSVPMixin):
     template_name = "main/homepage.html"
 
-    def get_non_main_meetings(self, num):
-        return (
-            Meeting.objects.filter(meeting_type__isnull=False)
-            .filter(when__gt=datetime.datetime.now() - datetime.timedelta(hours=6))
-            .order_by("when")[:num]
-        )
-
     def get_meeting(self):
         return (
-            Meeting.objects.filter(meeting_type__isnull=True)
+            Meeting.objects
             .filter(when__gt=datetime.datetime.now() - datetime.timedelta(hours=6))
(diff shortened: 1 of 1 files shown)


chipy_org/apps/main/views.py
```diff
@@ -36,10 +29,7 @@ def get_context_data(self, **kwargs):
         context = {}
         context.update(kwargs)
         context["IS_HOMEPAGE"] = True
-        context["other_meetings"] = self.get_non_main_meetings(num=3)
         context["featured_sponsor"] = Sponsor.featured_sponsor()
-        context["sponsor_groups"] = SponsorGroup.objects.prefetch_related("sponsors")
```
nickymarino, line 41: Were these removed intentionally? They seem unrelated to the __main__ vs other meeting fix 🤔

nickymarino (approved):
Lines 41-42 seem unrelated to this fix to me, but I might be wrong (see comment). Otherwise looks great!

chipy_org/apps/main/views.py
sophm256: Yes, because they're no longer needed in the redesigned homepage.

elmq0022 (approved):

PR #419, opened 2021-07-04
Add slack integration for job board

elmq0022:
This PR adds a custom Django command which can be scheduled to post the list of current approved jobs to slack.  Given we use Heroku I suggest we use their scheduler to post to the jobs channel daily in the early A.M.  Maybe 7AM?

### DIFF
--- a/chipy_org/apps/job_board/forms.py
+++ b/chipy_org/apps/job_board/forms.py
@@ -29,7 +29,9 @@ class JobPostForm(forms.ModelForm):
         ]
 
         help_texts = {
-            "affiliation": ("Is this posting affiliated with a 3rd party? Please select:"),  # pylint: disable=line-too-long
+            "affiliation": (
+                "Is this posting affiliated with a 3rd party? Please select:"
+            ),  # pylint: disable=line-too-long
         }
 
         widgets = {
--- a/chipy_org/apps/job_board/forms.py
+++ b/chipy_org/apps/job_board/forms.py
@@ -29,9 +29,7 @@ class JobPostForm(forms.ModelForm):
         ]
 
         help_texts = {
-            "affiliation": (
-                "Is this posting affiliated with a 3rd party? Please select:"
-            ),  # pylint: disable=line-too-long
+            "affiliation": ("Is this posting affiliated with a 3rd party? Please select:"),
         }
 
         widgets = {
--- a/chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
+++ b/chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
@@ -21,7 +21,6 @@ class Command(BaseCommand):
     """
 
     def handle(self, *args, **options):
-        print("These are the Active Postings on the Job Board", "\n")
         posts = JobPost.approved_and_active.all()
 
         if posts.count():
--- a/chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
+++ b/chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
@@ -1,4 +1,5 @@
 import json
+import logging
 
 import requests
 from django.conf import settings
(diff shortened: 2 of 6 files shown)


chipy_org/apps/job_board/forms.py
```diff
@@ -29,7 +29,9 @@ class Meta:
         ]
 
         help_texts = {
-            "affiliation": ("Is this posting affiliated with a 3rd party? Please select:"),  # pylint: disable=line-too-long
+            "affiliation": (
+                "Is this posting affiliated with a 3rd party? Please select:"
+            ),  # pylint: disable=line-too-long
```
raymondberg, line 34: Do you still need the line-too-long?

chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
```diff
@@ -0,0 +1,54 @@
+import json
+import logging
+import datetime
```
raymondberg, line 3: (non-blocking) Import sorts are always <3

chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
```diff
@@ -0,0 +1,54 @@
...
+            help="""
+            The desired weekdays to post represented an integer where Monday=0, Tuesday=1, ... , Saturday=5, and Sunday=6.
+            Posts to the slack channel will only occur on the selected weekdays.
+            """,
+        )
+
+    def handle(self, *args, **options):
+        posts = JobPost.approved_and_active.all()
+
+        if posts.count() and datetime.date.today().weekday() in options["weekdays"]:
```
raymondberg, line 38: Truly fine in the context of adding the code, but from a style perspective this is a little muddied with the concerns of the command. Ideally, I think we'd be well-served to put most of this into a Slack facade that handles renders and requests calls.

chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
```diff
@@ -0,0 +1,54 @@
...
+            slack_data = {"text": msg}
+
+            response = requests.post(
+                webhook_url,
+                data=json.dumps(slack_data),
+                headers={"Content-Type": "application/json"},
+            )
+
+            if response.status_code != 200:
+                logger.info("Failed to post to slack job channel.")
```
raymondberg, line 54: This should probably be a raised CommandError instead-of/as-well-as a log-entry: https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/#command-exceptions

chipy_org/settings.py
```diff
@@ -123,7 +127,7 @@ def env_list(key, defaults=None, delimiter=","):
 
 # Make this unique, and don't share it with anybody.
 SECRET_KEY = env_var("SECRET_KEY")
-
+DAY_TO_POST = int(env_var("DAY_TO_POST", 6))
```
raymondberg, line 130: Hmmm, what is this for? I'd push to make any settings like this human-readable. DoW math is notoriously wrong with either Sun=0 or Sun=1.

raymondberg (approved):
Cool to see this work!

chipy_org/settings.py
elmq0022: Leftover code will delete this.  Was going to do it as a env var, but moved it to a command argument.

chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
elmq0022: That's a reasonable idea and an opportunity for me to practice a design pattern.  Though I think I'll leave it until we add slack use case number two if that's cool.

chipy_org/apps/job_board/management/commands/post_approved_jobs_to_slack.py
elmq0022: :hamburger:

chipy_org/apps/job_board/forms.py
elmq0022: nope

PR #428, opened 2021-11-23
Add rsvp wait list

elmq0022:
Adding Wait List functionality and RSVP venue type - in person or virtual


.gitignore
```diff
@@ -73,3 +73,6 @@ venv/
 
 .vscode
 htmlcov
+
+Pipfile
+Pipfile.lock
```
raymondberg, line 78: What now? These shouldn't be ignored.

chipy_org/templates/shiny/_rsvp.html
```diff
@@ -12,8 +12,9 @@
         <p style="color: red">This event is no longer accepting registrations.</p>
     {% endif %}
     <p>
-        <strong>Current Attendance:</strong>
-        {{ next_meeting.number_rsvps}} Pythonista{{next_meeting.number_rsvps|pluralize}}
+        <strong>Current Attendance:</strong><br />
+        In Person Pythonistas: {{ next_meeting.number_in_person_rsvps}}<br />
+        Virtual Pythonistas: {{next_meeting.number_virtual_rsvps}}<br />
```
raymondberg, line 17: This might create some confusion if virtual or in-person doesn't make sense for some. Maybe recommend wrapping each with if statements.

chipy_org/apps/meetings/models.py
```diff
@@ -249,8 +300,13 @@ class RSVP(CommonModel):
...
     first_name = models.CharField(max_length=MAX_LENGTH, blank=True, null=True)
     email = models.EmailField(max_length=255, blank=True, null=True)
+
     meeting = models.ForeignKey(Meeting, on_delete=models.CASCADE)
-    response = models.CharField(max_length=1, choices=RSVP_CHOICES)
+
+    response = models.CharField(max_length=1, choices=RSVP_RESPONSE, default=YES)
+    status = models.CharField(max_length=1, choices=RSVP_STATUS, blank=True, null=True)
+    venue = models.CharField(max_length=1, choices=RSVP_VENUE, default=IN_PERSON)
+
```
raymondberg, line 309: My general thoughts on these is that I'm not a big fan of single character database values. We have a scale where optimizing for this isn't important (we pay per record way before we would pay per MB) and it makes intuiting things about the data more challenging.

chipy_org/apps/meetings/models.py
```diff
@@ -249,8 +300,13 @@ class RSVP(CommonModel):
     last_name = models.CharField(max_length=MAX_LENGTH, blank=True, null=True)
     first_name = models.CharField(max_length=MAX_LENGTH, blank=True, null=True)
     email = models.EmailField(max_length=255, blank=True, null=True)
+
     meeting = models.ForeignKey(Meeting, on_delete=models.CASCADE)
-    response = models.CharField(max_length=1, choices=RSVP_CHOICES)
+
+    response = models.CharField(max_length=1, choices=RSVP_RESPONSE, default=YES)
+    status = models.CharField(max_length=1, choices=RSVP_STATUS, blank=True, null=True)
+    venue = models.CharField(max_length=1, choices=RSVP_VENUE, default=IN_PERSON)
```
raymondberg, line 308: Would this be clearer with `attendance_type` or `attendance_method`? We already have a domain concept for `venue` that this could be easily confused with. 

(Case study: I did confuse it while reading this PR)

raymondberg: Also, this feels like it could be merged with response. The types of response seem to be `attending_in_person, attending_virtually, not_attending` rather than decoupling these two data points.

chipy_org/apps/meetings/models.py
```diff
@@ -117,6 +121,11 @@ def __str__(self):
     )
     description = tinymce_models.HTMLField(blank=True, null=True)
 
+    in_person_capacity = models.PositiveSmallIntegerField(null=False)
```
raymondberg, line 124: Having done this for a long time I'm torn on the in_person_capacity not having a default. I think that might make sense...but honestly setting it to something low might make more sense. Like setting it to 50 or null by default would reduce a barrier to creating these events. You might publish events 1 week later because it will take time to figure out what the capacity is, which I would think is worse than just setting a number that is a small default and raising it over time.

That said, I can see the case to leave it like this.

chipy_org/apps/meetings/models.py
```diff
@@ -131,11 +140,37 @@ def is_future(self):
...
+
+    def number_in_person_rsvps(self):
+        return self.rsvp_set.filter(
+            response=RSVP.YES, status=RSVP.ACCEPTED, venue=RSVP.IN_PERSON
+        ).count()
+
+    def number_virtual_rsvps(self):
+        return self.rsvp_set.filter(
+            response=RSVP.YES, status=RSVP.ACCEPTED, venue=RSVP.VIRTUAL
+        ).count()
```
raymondberg, line 154: That you created helper methods based around [the concept I talk about here](https://github.com/chicagopython/chipy.org/pull/428/files#r754746545) is a sign that you should merge these values down. If your code base is scaling with pairs of database options then it makes sense to merge them in the database as well.

chipy_org/apps/meetings/models.py
```diff
@@ -236,9 +271,25 @@ def __str__(self):
...
+
+    ACCEPTED = "A"
+    DECLINED = "D"
+    WAIT_LISTED = "W"
+
+    RSVP_STATUS = (
+        (ACCEPTED, "Accepted"),
+        (DECLINED, "Declined"),
+        (WAIT_LISTED, "Wait Listed"),
+    )
```
raymondberg, line 286: The RSVP statuses seem more business logic than finite state machine (which is what you want, I think). Much more common to go with a pattern like: 

> pending - no decision made
 confirmed - rsvp request can be fulfilled
 rejected - rsvp request cannot be fulfilled
 waitlisted - rsvp request could not be fulfilled on prior attempt, may be fulfilled in future

Pending might be YAGNI, but if you move to a worker RSVP model (single worker to avoid race conditions) then you'll be ready out of the gate.

chipy_org/apps/meetings/models.py
```diff
@@ -236,9 +271,25 @@ def __str__(self):
 
 class RSVP(CommonModel):
 
-    RSVP_CHOICES = (
-        ("Y", "Yes"),
-        ("N", "No"),
+    YES = "Y"
+    NO = "N"
+    RSVP_RESPONSE = ((YES, "Yes"), (NO, "No"))
```
raymondberg, line 276: Would recommend naming these as plural of the name of the name of the field, i.e. `RESPONSES` or whatever you wind-up with. 

I also recommend doing nested classes for this: 

```
class RSVP(CommonModel):
  class Response:
    YES = "yes"
    NO = "no"
    ALL = [YES, NO]
    CHOICE_LIST = [(YES, "YES"), (NO, "NO")]
```

chipy_org/apps/meetings/tests/test_rsvps.py
```diff
@@ -109,7 +114,9 @@ def test_unique_rsvp(self):
         Tests the uniqueness constraints on the rsvp model
         """
         test_venue = Venue.objects.create(name="Test")
-        meeting = Meeting.objects.create(when=datetime.date.today(), where=test_venue)
+        meeting = Meeting.objects.create(
+            when=datetime.date.today(), where=test_venue, in_person_capacity=5
```
raymondberg, line 118: It seems like we're using `where` too many places. As you work in here feel free to remove anything that doesn't need to assert things about `where`.

chipy_org/apps/meetings/templates/meetings/emails/rsvp_email.html
```diff
@@ -1 +1,24 @@
-To change your rsvp, click <a href="http://{{ site.domain }}{% url 'update_rsvp_with_key' key %}">here</a>
+Your {{ rsvp.get_venue_display }} RSVP for {{ rsvp.meeting }} has been {{ rsvp.get_status_display }}.
+
+{% if rsvp.status == "A" %}
```
raymondberg, line 3: Seems like this should have more conditions based on attendance. If they are attending in-person this makes sense, but there's a condition missing here for the others.

This is all about de-incentivizing folks from taking slots, so the check is rather complicated. Are they going in-person? If not, does it have a virtual event limit? If either of the previous questions then display. And waitlists fill up too, so waitlists for either might be included.

raymondberg (commented):
I finished my first pass. I would think about the data model a bit to simplify your logic and add a few unit-tested properties on the RSVP model to make your email logic easier. Think `rsvp.contributes_to_capacity`

chipy_org/templates/shiny/_rsvp.html
elmq0022: Good point thanks.

chipy_org/apps/meetings/models.py
elmq0022: fair enough - the single chars do become easier to implement in Django 3.2 - but you do run into nested classes ...

.gitignore
elmq0022: Correct - will remove.

chipy_org/apps/meetings/models.py
elmq0022: I didn't like "venue"  when I wrote it.  I like attendance_type much better.

chipy_org/apps/meetings/models.py
elmq0022: I could go either way on the default size.  I think the bigger issue is changing the in_person_capacity after the fact regardless.  I need to capture a post_save signal on meeting and adjust the wait list based on the new capacity.

chipy_org/apps/meetings/models.py
elmq0022: I agree on merging the response and status. Would you just migrate all the old "YES" values to IN_PERSON?  Probably doesn't matter as we have never done analysis on old RSVP data.

chipy_org/apps/meetings/models.py
elmq0022: Good call.

chipy_org/apps/meetings/tests/test_rsvps.py
elmq0022: :+1:

chipy_org/apps/meetings/templates/meetings/emails/rsvp_email.html
elmq0022: I read this a couple time.  I think I get it.  

Basically you're saying we should **not** de-incentivized people if the meeting has no capacity limit.  For example, we would probably not set a limit for virtual meeting so, telling them to update their RSVP really isn't necessary.  

The wait lists won't fill up.  Anyone not accepted will be wait listed.  Though they could be so far down the list they practically would never get a spot.

chipy_org/apps/meetings/models.py
raymondberg: Yeah, I think the right answer here is to figure out what stats we care about from prior attendance and create a single record with that information and delete old RSVPs. Rationale: this is our largest row generator and we hit costs thresholds by row count.

But you could safely do everything pre 3/2021 as `attending_in_person` and everything since then as `attending_virtually`

chipy_org/apps/meetings/templates/meetings/emails/rsvp_email.html
raymondberg: But if someone decides to never attend and they are on the waitlist shouldn't they update their status when it's convenient rather than risking them not doing it once they get a slot?

The idea is that you want folks to take action if their action will improve the experience for themselves or others. The simple answer here is that action isn't based on whether they are `confirmed` (A) or `waitlisted` but rather whether  their reservation would block another's involvement on the night.

chipy_org/apps/meetings/models.py
elmq0022: I made it 50 chars with a full description.

chipy_org/apps/meetings/models.py
elmq0022: merged to response only.

chipy_org/apps/meetings/models.py
elmq0022: merged

chipy_org/apps/meetings/models.py
elmq0022: Done.

chipy_org/apps/meetings/templates/meetings/emails/rsvp_email.html
elmq0022: Well that is basically everyone who has not declined their RSVP.  So, I'll just remind folks to please update their status if they change their mind regardless.  

The thing to do here is to estimate how many no shows we typically have and then overbook accordingly.   I don't have a clue what that percentage would be or how to gauge the enthusiasm for a first in-person chipy meeting in two years.

chipy_org/apps/meetings/templates/meetings/emails/rsvp_email.html
```diff
@@ -1 +1,25 @@
...
+    {% else %}
+    go to <a href="http://{{ site.domain }}">chipy.org</a>
+    {% endif %}
+
+to change your response to "DECLINED". 
+This will help those who still wish to attend receive a confirmation.</p>
+{% endif %}
+
+{% if rsvp.status == "wait listed" %}
+<p>If your reservation is confirmed you will be notified by email.</p>
```
raymondberg, line 18: I would move this to the top of the declined block, say line 5?

```suggestion
<p>You will be notified by email if a spot becomes available and you are taken off the wait list.</p>
```

chipy_org/apps/meetings/forms.py
```diff
@@ -76,9 +76,26 @@ def __init__(self, request, *args, **kwargs):
         if self.instance.pk:
             del self.fields["email"]
 
+        meeting = kwargs.get("initial").get("meeting", None)
```
raymondberg, line 79: This access pattern surprises me. Can you walk through why this is needed now and why we can't rely on `super()` to get some of this data?

chipy_org/apps/meetings/models.py
```diff
@@ -276,17 +331,85 @@ def clean(self):
...
     @property
     def full_name(self):
         return self.first_name + " " + self.last_name
 
     def __str__(self):
         return f"{self.meeting}: {self.full_name}"
+
+
+@receiver(post_save, sender=RSVP)
+def rsvp_post_save(sender, instance, **kwargs):
```
raymondberg, line 371: This method is a great case for us getting that background worker in place. Imagining a case where 100 people were on the backlog post resize could trigger 99 queries and 99 saves and 99 emails in serial.

chipy_org/apps/meetings/models.py
```diff
@@ -276,17 +331,85 @@ def clean(self):
...
+    # don't updated anything if registration is closed
+    # if not meeting.can_register():
+    #     return
+
+    # send an email to the user by email
+    if instance.email:
+        send_rsvp_email(instance)
+
+    if meeting.has_in_person_capacity():
+        first_on_in_person_wait_list = (
```
raymondberg, line 388: I would follow the "fat models" approach and move this into the meeting object. `meeting.allow_waitlisted_individuals(1)` or something like that seems reasonable.

It would help decouple your hook code from the logic to manage the RSVPs.

raymondberg: That said, this is clearly functional :D

raymondberg (approved):
Added a few more comments, but this really looks good.

PR #429, opened 2021-11-23
Use Pipenv to Manage requirements

raymondberg:
## Problem

We don't have a formal locking system for requirements so there are loads more packages installed
during a `pip install` than we track. What does this mean? Well there are [lots of articles](https://jonathanmeier.io/advantages-of-pipfile-lock-over-requirements-txt/)
talking [about why](https://lincolnloop.com/blog/python-dependency-locking-pip-tools/) you should [use a locking](https://modelpredict.com/wht-requirements-txt-is-not-enough) solution [for your projects](https://towardsdatascience.com/solving-dependency-management-in-python-with-poetry-165b92069e9d)

## Solution

Let's do that! We can pick between the three front-runners: Pipenv, Poetry, and Pip-tools. But the importance is
just getting something in place sooner than later. Folks can absolutely open PRs to recommend other approaches,
but Pipenv is quick and easy.

Note, I updated to 3.8.10 as part of this because [it was easier to work on my Mac M1 (see macOS users)](https://www.python.org/downloads/release/python-3810/).


docker/Dockerfile
```diff
@@ -22,9 +23,9 @@ WORKDIR ${SITE_DIR}
 RUN install -d proj/ var/log/ htdocs/ htdocs/static/
 
 # Install python packages
-ADD requirements.txt ${SITE_DIR}requirements.txt
-ADD requirements_test.txt ${SITE_DIR}requirements_test.txt
-RUN pip install -r ${SITE_DIR}requirements_test.txt
+ADD Pipfile ${SITE_DIR}Pipfile
+ADD Pipfile.lock ${SITE_DIR}
```
elmq0022, line 27: `ADD Pipfile.lock ${SITE_DIR}` or `ADD Pipfile.lock ${SITE_DIR}Pipfile.lock`

raymondberg:
```suggestion
ADD Pipfile.lock ${SITE_DIR}Pipfile.lock
```

raymondberg:
Merged with #431

Issue #427, opened 2021-11-21
Sponsor Logos are not sized to width on mobile layout


elmq0022:
Need some CSS or some other method limiting the image size to the mobile layout.  

![Screenshot_20211121-071101](https://user-images.githubusercontent.com/3450848/142763399-9b2b7045-0f6a-44de-bbea-d442851c698c.jpg)

elmq0022:
@spark-c is this something you could pick up?   You seem pretty good with CSS.

spark-c:
Sure thing!

elmq0022:
Thanks! I would really appreciate that.  It's important that the site look
good especially for our sponsors.

PR #430, opened 2021-11-23
Responsive sponsor_detail logo

spark-c:
Should resolve issue #427 

I went poking around in bootstrap and found a bs-class that will handle making the logo image responsive, documentation found [here](https://getbootstrap.com/docs/4.0/content/images/#responsive-images).

The localhost version of the site/db doesn't have logos to test upon, so I did a brief experiment on the live site with browser devtools. Adding the class `img-fluid` to the appropriate img tag in the sponsor_detail.html template made the image responsive and it seemed to scale correctly (tried for screen sizes up to 1920x1080).

Here's a little video:

https://user-images.githubusercontent.com/51402465/143141252-2104430c-df23-4b9a-aa04-5f78c1af2370.mp4

I'm not quite sure how to get a sample image into the dev db for testing it properly, or if that would be the correct route -- however I'm packing+moving house at the moment, and don't have the time yet to figure it out properly.

Anyway, this PR just adds the `img-fluid` class to the img tag in sponsor_detail.html

:)

### DIFF
--- a/chipy_org/apps/sponsors/templates/sponsors/sponsor_detail.html
+++ b/chipy_org/apps/sponsors/templates/sponsors/sponsor_detail.html
@@ -11,7 +11,7 @@
 {% thumbnail sponsor.logo "400" crop="center" as im %}
   <div>
     <a href="{{ sponsor.url }}">
-    <img src="{{ im.url }}" alt="{{ sponsor.name }}"></a>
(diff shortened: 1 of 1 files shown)


elmq0022 (approved):
Awesome!

elmq0022:
I've deployed this.  It looks so much better now.  

Also - yes you can go through the admin to create a sponsor and add an image.  We have a make dev data command but I think I only added the names there.  We could add a couple images as test data if you like.

PR #431, opened 2021-11-24
Pipenv with local dev

raymondberg:
## Problem

I want to do a bunch of iterations on the Pipfile but doing that in Docker is very, very tedious.

## Solution

Allow for folks to run Django locally...with caveats that it's unsupported.


raymondberg:
Relies heavily on commits from #429 , recommend merging that before reviewing.

README.md
```diff
@@ -164,3 +164,13 @@ regardless you can deploy a feature branch with the following command:
...
+
+
+## Alternate Local Development
+
+You can use `make up-services` to start the background services
+locally. Then, if you copy `docker/docker.env` to `.env` you can
+use pipenv to run the service locally: `pipenv run python manage.py runserver`.
+
+If you develop locally you are on your own! All contributions must work in
+docker.
```
elmq0022, line 176: Maybe add something like: 

Of course, you will need pipenv and the correct python version installed locally.

elmq0022 (approved):
I really like this one, but I feel a little guilty about that.  I look forward to using standard shell commands and such.  

See my comment about the README.md.  Though if someone doesn't know how to get pipenv and the correct python version they should maybe stick to docker.

README.md
raymondberg:
```suggestion
docker. This path requires Python and Pipenv installation knowledge.
```

docker/Dockerfile
```diff
@@ -22,9 +23,9 @@ WORKDIR ${SITE_DIR}
 RUN install -d proj/ var/log/ htdocs/ htdocs/static/
 
 # Install python packages
-ADD requirements.txt ${SITE_DIR}requirements.txt
-ADD requirements_test.txt ${SITE_DIR}requirements_test.txt
-RUN pip install -r ${SITE_DIR}requirements_test.txt
+ADD Pipfile ${SITE_DIR}Pipfile
+ADD Pipfile.lock ${SITE_DIR}
```
raymondberg, line 27:
```suggestion
ADD Pipfile.lock ${SITE_DIR}Pipfile.lock
```

Issue #433, opened 2021-11-24
Upgrade Django Social ahead of Django 4.0


raymondberg:
We have deprecation waning and need to get on a newish version of Django-social. 

Check by running django with `python -Wd ...` to see whether you've fixed the issue.

Issue #432, opened 2021-11-24
Switch from norecaptcha-recaptcha to django-recaptcha


raymondberg:
We used an unmaintained fork from Imaginary Landscape to get us through the recaptcha v2 conversion years ago, but it has no updates and cannot support Django 4.0. 

Switch to the standard django-recaptcha library and get our captchas working!

Check by running django with `python -Wd ...` to see whether you've fixed the issue.

elmq0022:
hmmmm ... seems unmaintained 

https://github.com/praekelt/django-recaptcha/issues/244

will added it because it's easy and hope someone updates it or forks it.

elmq0022:
we could move to simple captcha as well.

PR #434, opened 2021-11-24
Pipenv upgrade django to 3 (and prepare for 4)

raymondberg:
## Problem

Django 2.2 is already out of mainstream support and will soon be out of Extended Support.

## Solution

We should upgrade to the latest version...and get ready for Django 4 which will come out soon! 

### What is covered

Perform all updates up to 3.2. This has been tested locally and all of the key features seem to work.

I ran into some issues with later updates of some packages, so I did whatever was minimally required 
to stop the Deprecation warnings from running in `python -Wd` mode.

### What is not covered

Upgrading django_social is always a pain, so I skipped all pre-4.0 warnings there. https://github.com/chicagopython/chipy.org/issues/433

I also skipped norecaptcha-recaptcha https://github.com/chicagopython/chipy.org/issues/432

There are other deprecation warnings in tests that existed before I started but also relate to the two above. I'd like to get those two issues fixed before trying to clear up test deprecations.

This relies heavily on other PRs needing merging #429 and #431


docker/Dockerfile
```diff
@@ -22,9 +23,9 @@ WORKDIR ${SITE_DIR}
 RUN install -d proj/ var/log/ htdocs/ htdocs/static/
 
 # Install python packages
-ADD requirements.txt ${SITE_DIR}requirements.txt
-ADD requirements_test.txt ${SITE_DIR}requirements_test.txt
-RUN pip install -r ${SITE_DIR}requirements_test.txt
+ADD Pipfile ${SITE_DIR}Pipfile
+ADD Pipfile.lock ${SITE_DIR}
```
elmq0022, line 27: Quick question - Should the new line 27 be `Pipfile.lock ${SITE_DIR}Pipfile.lock`?  Looks odd compared to Pipfile and the old requirements.txt  and the requirements_test.txt.

elmq0022 (commented):
Wow.  That was fast @raymondberg.  

So you didn't bump into any issues with incompatible packages?

docker/Dockerfile
raymondberg: I'll fix this with a rebase

raymondberg:
Nothing hard on this upgrade as far as I can tell. TinyMCE broke on later versions, but I got it up to 4.0 compatiblity

elmq0022 (approved):
Looks great! Thanks.

PR #450, opened 2021-12-19
updates django social auth

elmq0022:
@raymondberg the update seems to be relatively painless this time unless I'm missing something.  Please take a look.  Thanks!


elmq0022:
fixes #433

PR #489, opened 2023-04-18
Replace django-flatpages-tinymce with django3-...

raymondberg:
Our builds stopped working because someone has apparently wiped
django-flatpages-tinymce from the face of the Earth.

There's a port that drops support for Django2. And isn't it great that
we are on Django3?  Yes! Always bump versions. And since we use Pipenv
and pin on each version rebuild...there's not much reason to pin these
lesser library requirements. If someone needs to do it again because
something breaks they can absolutely do it, but rebuilding the lockfile
should come with comprehensive testing.

In this case I just loaded it up locally and saw that the MCE header was still on the textarea

image


raymondberg:
Meh, I missed and this seems to be fixed on main.

PR #493, opened 2023-06-13
Refactor - Make new app for talks

raymondberg:
## What

We want to do a few things to improve the way that talks and topics are
managed. To do this, it's going to be cleaner to split out the topics/talks
from RSVPs and meetings and venues in the already cluttered meetings app.

## Why?

Talks have a relationship to meetings, but it's not central to its identity.
When talks are created, they are created without a reference to meetings and
they could be moved between meetings or never get assigned.

## Why not the models?

Model moving between apps is a bit trickier. We may want to make new tables and
migrate the data manually rather than doing some rather painful refactoring.

## What next?

Well, we should move the models, but maybe not into things named as they are
today. Topics are really "TalkProposals" and there are lots of other models that
could exist: Theme/Topic, Talk/Topic Voting, Speakers, Nominations, Speaker
Feedback, Videos, Analytics ...

All these things are worth diving into via either by moving, creating new, or
creating replicated tables.

## Demo

image

image

image


eevelweezel (approved):
I'm approving in the interest of expediting the fix for notifications.

PR #494, opened 2023-06-15
Replace hardcoded emails with configurable ones

raymondberg:
We need to get email working again, but some of our emails are hardcoded to an address that doesn't work. Let's get them all using the existing setting.


JoeJasinski:
Looks good

PR #495, opened 2023-06-16
Rewrite email sending to stop sending from variable sources

raymondberg:
Sendgrid turned on authorized senders and the email address(es) we were using
stopped working. We set up a new email address to get this working, but we need
to remove every occurrence of setting the "from" email on the message, since we
are now locked in.

We had 3 different implementations using the low-level Django message interface
which meant email senders needed to know how email settings worked. This PR
moves us to a single implementation that obscures the from setting.


raymondberg:
Merging in the spirit of fixing for @eevelweezel

PR #508, opened 2023-09-19
add phone to topic submission

JoeJasinski:
I believe this will make it so that talk submitters need to enter a contact phone number when they submit a talk. I'm hoping that this will make it easier to contact a speaker if we need to reach them on the day of a talk. Let me know what you think.


raymondberg (approved):
Love it!

PR #509, opened 2023-09-19
expose speaker emails in the admin interface 

JoeJasinski:
When trying to email speakers in the past, I've found it difficult to get their email addresses out of the admin interface. I usually have to view source on the speaker multi-select box on the meeting page in order to copy out their email addresses.

Instead, I added a new readonly field to the admin Meeting page with a list of email addresses in RFC 5322 "mailbox" format. This allows us to much more easily identify the emails of the speakers. 

Let me know what you think.


raymondberg (approved):
Love it!

PR #515, opened 2024-01-09
Feature/reviewer

CarlFK:
Reviewer role is someone who gets the dubious privilege to watch the video from start to finish and decide if it is good enough or needs fixing. 
more:   https://github.com/CarlFK/veyepar/wiki/Reviewer

This PR should be enough so that an admin can put the data in the database using the admin UI, and the export exports it.

sad confession: it is untested.  I hope it works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ docker.env
 htdocs/static/
 htdocs/media/
 chipy_org/mediafiles/
+venv/
--- a/chipy_org/apps/main/views.py
+++ b/chipy_org/apps/main/views.py
@@ -3,7 +3,7 @@ import datetime
 import sys
 import traceback
 
-from django.http import HttpResponseServerError
+from django.http import HttpResponse, HttpResponseServerError
 from django.template import loader, Context
 from django.views.generic import TemplateView
 from chipy_org.apps.meetings.models import Meeting, RSVP
@@ -69,3 +69,6 @@ def custom_500(request):
         'exception_value': value,
         'value': type,
         'tb': traceback.format_exception(type, value, tb)})))
+
+def customer_404(request):
+    return HttpResponse('<h1>404 - Page Not Found</h1>', status_code=404)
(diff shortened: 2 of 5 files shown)


raymondberg:
From an organizational perspective, it feels like this would be easier to manage if this were always the same person on our side rather than putting it in the hands of many (3 speakers x 12 months = 36 reviewers per year). Is there a specific reason to have this per-talk?

CarlFK:
it is the crowd sourcing model.  spread the work out, and to people that want to do the work for reasons other than "someone needs to do this work."  

from https://github.com/CarlFK/veyepar/wiki/Reviewer

**More fluff: I want to emphasize that each reviewer should only be responsible for one talk. The reviewer only has one job: to catch any mistakes that might happen to this one video. This one video should get their complete attention. "All the videos" is an unreasonable task as it requires discovering what the actual set of videos is....**

I can CC one or more people on all of this, and if someone else gets to it first, that's fine.  There shouldn't be the problem of 2 people doing the same 'work' because it shouldn't be work, it should be a bunch of people watching the same video, which is kinda the goal of making these videos: 1000's of people watching them. 

@WeepingClown13 (Ananthu, Debian video team) has offered to watch all the videos.  I like this, but I also don't.  Someday it will be a task that they are obligated to do that they don't really want to do, and they are not getting any benefit, other than having access to videos a day earlier, which isn't really 'value'.  It's barely prestige.  I feel it is taking advantage of someone's generosity, and that's something we should all try to avoid.

WeepingClown13:
Both makes sense to me, but yes, having more people always pays out. For one the result doesn't just depend on a single person that way, and the person will not find it a "job" either. To backup what Carl said, I agreed to review one of the recent videos and then got tied to the bed with a fever, only to find out later someone else already did the job and approved the video, which worked out in everyone's favor.

raymondberg:
Edit: Alternatively, if it's just email we can also send it to a distribution list. We have a speakers-guild email distro that we coordinate speakers on already. The same group could be responsible for reviewing. Let me know if that's preferable.

CarlFK:
come to holiday club tomorrow wednesday, 6-9pm.  we will get this figured out.  and auth to sensitive info too.

raymondberg:
This was resolved with https://github.com/chicagopython/chipy.org/pull/518

PR #517, opened 2024-01-11
Stop using circleci

raymondberg:
I have to move some code and the tests all pass locally and CircleCI has changed its interface locking me out of the project. We can revisit, but this isn't workable now.


eevelweezel:
Approving, with the caveat that we should figure out a CI replacement.

PR #519, opened 2024-01-11
Add presenter email to protected api dump

raymondberg:
Now that the route is protected, we can more simply expose presenter email


JoeJasinski (approved):
Looks great

Issue #551, opened 2024-11-12
Add Third Party Build System to Manage Dependencies.  


elmq0022:
I miss pipenv already. Can we use something else? I'd be in favor of using UV as it's really really fast.  I don't have a lot of time with it though and I can't say for sure that it's really really good.  

Any other players in this space that people like?

eevelweezel:
I've used hatch, and it ultimately depends on UV. I'd be in favor of using
either.

Sorry for evicting pipenv without consulting others, but after 3 hours, I
still couldn't resolve that conflict. I'm happy to assist with this, if I
can do so without getting in the way.

elmq0022:
No worries about pipenv. I'd like to give uv a go.  I could use the experience using uv in another project.

elmq0022:
So Heroku says ...

I only like requirements.txt, Pipfile.lock, or poetry.lock for deploys.  pyproject.toml be damned! 

My brain immediately jumped to poetry, but pip-tools might be fine as well.

elmq0022:
https://github.com/jazzband/pip-tools

PR #556, opened 2024-11-15
Django 5.1 and Pip-Tools

elmq0022:
pip-tools for the win!

I tried poetry and :-1:  wasted a bunch of time.  This may have been a skill issue, but if so, poetry is too damn hard for what it's doing, imho.


elmq0022:
closes #551

PR #558, opened 2024-11-18
Use post method for logout

### LINKED ISSUE #557: Logout Url Not Working
After updating to Django 5.1 the logout URL for social auth is not working.  I'm not sure if this was an issue before or not.  

Logging out from the Admin is fine.

elmq0022:
Fixes #557 

The anchor tags <a /> are html get request by default.  The logout view requires a post method to logout.  

I wrapped the link in a small form, added the crsf token and put the logout text in a button.   

The css should be adjusted though so the button doesn't have an outline ... :frowning_face:


elmq0022:
Well `border: none` will probably clean up the css. That or `outline: none`.  Seems like that second one isn't recommended for access though.

eevelweezel (approved):

elmq0022:
Example of the odd CSS when switching this to a button.  

image

elmq0022:
CSS fixed!

Issue #535, opened 2024-10-24
security warnings for  django-ckeditor


eevelweezel:
(ckeditor.W001) django-ckeditor bundles CKEditor 4.22.1 which isn't supported anmyore and which does have unfixed security issues, see for example https://ckeditor.com/cke4/release/CKEditor-4.24.0-LTS . You should consider strongly switching to a different editor (maybe CKEditor 5 respectively django-ckeditor-5 after checking whether the CKEditor 5 license terms work for you) or switch to the non-free CKEditor 4 LTS package. See https://ckeditor.com/ckeditor-4-support/ for more on this. (Note! This notice has been added by the django-ckeditor developers and we are not affiliated with CKSource and were not involved in the licensing change, so please refrain from complaining to us. Thanks.)

elmq0022:
Cripes I know we had two different rich text editors.  Looks like I decided to yank out the wrong one.  

Maybe we find an alternative for creditor.

elmq0022:
I took a look at the upgrade path to django-ckeditor-5.  It's not plug and play.  The names of fields and widgets changed so and upgrade needs db migrations.  :roll_eyes: 

Do we want to save the associated data? I would think so. Do I want to upgrade to ckeditor-5.  Not Really.  

Personally I'd like a widget for the models, sure, but a standard text field or text area would be fine.  Maybe some JS to change said field into a rich editor.

eevelweezel:
Is django-tinymce an option?

elmq0022:
Yes.  I was thinking about that.  Editor is bundled now and MIT on the django integration.  LGPL on the editor itself.  

I'm super disappointed with this upgrade path.

eevelweezel:
Ooh, or this? https://github.com/django-cms/djangocms-text

elmq0022:
It's shiny for sure, but doesn't look like it's been around long and not many stars.

elmq0022:
The more annoying part is that we have custom admin widgets with this django-ckeditor and the migration of the data.

elmq0022:
I will move us to tinymce, but just the widget.  All the dumb use of the ckeditor model field will be replaced to a models.TextField.

1. Create new TextFields to replace the ckeditor model fields.  
2. Migrated to the data to the new field
3. Forms will be updated to tinymce widget be it in the admin and the user facing site
4. Remove the field from the model
5. Migrate to delete the associated column from the db
6. Rename the fields back to the original name
7. Squash all the migrations
[...]

PR #560, opened 2024-11-20
Move to tinymce

elmq0022:
Starts to address #535 - will migrate to tinymce


eevelweezel (approved):
Looks good, thank you!

PR #568, opened 2024-11-28
Ruff life v 2.0

### LINKED ISSUE #552: Replace Black and Isort with Ruff
ruff replaces both of these tools and is much much faster

I don't care about the actual format so much as long as it's consistent and I don't have to wait the tool to get the job done.

eevelweezel:
Fix the config for ruff and pre-commit.ci. 

Closes #552.

Apologies for the duplication, cleaning up that problematic commit auto-closed my last PR.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.7.3
+  hooks:
+    # Run the linter.
+    - id: ruff
+      args: [ --fix ]
+    # Run the formatter.
+    - id: ruff-format
+repos:
```
elmq0022, line 11: Pretty sure ruff can sort imports.

elmq0022 (approved):

.pre-commit-config.yaml
eevelweezel: Alright, updated.

eevelweezel: It can, but so far, I don't like what it's doing when we trigger it with pre-commit.  The "way" is to run a filtered check, and it doesn't seem confined to the diff. I'm going to revert these last few commits & stick with isort for now.

eevelweezel: Created #569 to address this in the future.

PR #570, opened 2024-12-04
adding upcoming meetings back to main page and DRY queryset

JoeJasinski:
As a user of the website, I want to easily find the __main__ meeting (or other upcoming events) when a SIG appears directly before the __main__ meeting and occupies the only "slot" available. 

I'm adding back the upcoming meetings section that disappeared after the redesign. I've also done some DRY work on the queries that were related to the meeting selection.

image


elmq0022 (approved):

pyproject.toml
```diff
@@ -55,7 +55,13 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
+    "black",
+    "flake8",
+    "freezegun",
+    "isort",
```
elmq0022, line 61: I thought we were going with ruff for formatting and linting after @eevelweezel last PR.

pyproject.toml
```diff
@@ -55,7 +55,13 @@ dependencies = [
...
     "pre-commit",
+    "pylint",
+    "pylint-django",
```
elmq0022, line 64: same comment as above.

pyproject.toml
JoeJasinski: Not sure, I was getting a merge conflict when I pulled in the main branch and tried to resolve it as best I could (though I just branched off of main today). Did I merge wrong?

pyproject.toml
JoeJasinski: Probably the same issue as above. What should it be set to?

JoeJasinski: I'll try copying down the pyproject.toml from main, add my new dev dependency, and re-gen the lock files. 

I'm really confused though - I just pulled from main, so I don't know why I got all of those merge conflicts.

elmq0022: https://github.com/chicagopython/chipy.org/blob/main/pyproject.toml#L57

elmq0022: yeah that's weird to see all those merge conflicts.

JoeJasinski: ok, I tried that strategy and re-pushed. How does it look?

elmq0022: Lgtm

pyproject.toml
elmq0022: Upstream or origin 🤔

JoeJasinski: I thought I pulled main from chipy.org official project (and then pushed up that main to my main branch in my local repo). Something must have gotten messed up with that. Anyway, the diff in the merge looks like it now only includes my changes.

Issue #467, opened 2022-03-20
As an Administrator I want Django Recaptcha for Captchas


elmq0022:
The django recaptcha project was dropped by [Praekelt Consulting](https://github.com/praekelt/django-recaptcha), but is now continued by torchbox, see https://pypi.org/project/django-recaptcha/.   

We moved to django-simple-captcha, but it's ability to reduce spam from bots has been significantly less performant.  

I would like to move back to django-recaptcha to reduce the amount of spam email we admins receive via chipy.org.

PR #575, opened 2025-01-09
Use django recaptcha instead of simple recaptcha

elmq0022:
Typing the values for the simple captcha is a pain.  Django-recaptcha is supported again and our keys are still in the inventories.  This PR will put us back to using google's simple checkbox as a captcha.


eevelweezel (approved):

elmq0022:
address #467

elmq0022:
thanks @eevelweezel for the quick review.  I'll probably deploy after this weeks meeting just in case the api keys are busted.

Issue #577, opened 2025-01-10
Contact Form Appears to be Broken


elmq0022:
Sending an email via the contact form seems to be broken.  We should determine if that's the case and fix it if necessary.

eevelweezel:
reopening, as it appears to still be broken with social auth.  We may need to reconsider how we're testing.

PR #578, opened 2025-01-10
fixes the contact form and related view test.

elmq0022:
I think it's fixed.  At least it's not more broken.


elmq0022:
#577 is closer to fixed - still seeing issues with in the sendgrid dashboard.  It's likely this is no longer a code issue.

PR #580, opened 2025-01-10
updates README.md

elmq0022:
This PR: 

- updates the README to include details on how to use `pip-tools`
- Fixes the pre-commit yaml so both `ruff` and `isort` run (only `isort` was running)
- Updates the Makefile to use `ruff` for linting and formatting
- removes references to tagging a release; no one has tagged a release since 2021  : P

If someone want to add a hook or something to make tagging happen automatically that would be great, but I don't think we're going to get traction manually.


elmq0022:
Prior
image

Now 
image

eevelweezel (approved):

PR #582, opened 2025-01-12
add nh3 filter and replace bleach

### LINKED ISSUE #574: Replace bleach with html-sanitizer or similar
Chipy.org's html sanitizer, bleach, has been deprecated.  We should probably replace it with [html-sanitizer](https://github.com/matthiask/html-sanitizer) or something similar like [nh3](https://nh3.readthedocs.io/en/latest/), if you're a cool kid that likes oxidized python.

eevelweezel:
Adds a custom filter loosely based on django-nh3, but compatible with python3.13, and only implementing features we're using. 

Closes #574

### DIFF
--- a/chipy_org/libs/tests/test_nh3_templatetags.py
+++ b/chipy_org/libs/tests/test_nh3_templatetags.py
@@ -1,0 +1,41 @@
+from django.template import Context, Template
+from django.test import TestCase
+
+
+class TestNH3Templates(TestCase):
+    """Test template tags"""
+
+    def test_bleaching(self):
+        """Test that unsafe tags are sanitised"""
+        context = Context(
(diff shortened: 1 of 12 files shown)


chipy_org/libs/tests/test_nh3_templatetags.py
```diff
@@ -0,0 +1,41 @@
+from django.template import Context, Template
+from django.test import TestCase
```
elmq0022, line 2: We do have pytest django is installed, if you don't want to rock the old school unit tests framework.  I don't have a strong preference though.  I am just happy to see some tests. 😊

elmq0022:
Best branch name ever!  Approved 👍

elmq0022 (approved):

Issue #573, opened 2024-12-20
Add a way to mark a job post as filled


JoeJasinski:
As a job poster, I want to indicate if a job that I posted has been filled, so that I can remove the posting but demonstrate that it has been filled. 

We should add a column to the job post table to indicate if a job post has been filled and also a date indicating the filled date. The date could default to the current date. 

Acceptance Criteria: 
- Modify the model and forms to expose a filled checkbox (or similar) and date field to indicate filled.
   - Filter the job posting to filter out filled jobs
- A job poster should be able to mark a job as filled on the job update view. 
- (optional) An anonymous viewer should be able to see a list of unfilled and filled jobs.

eevelweezel:
I like the idea of tracking filled posts, but I'm wondering why we'd leave
filled posts visible?

I guess visible with a "filled" status isn't the behavior I would expect.

elmq0022:
I'm with heather on the above.  As a job seeker, if the position is filled I don't want to waste my time looking at it.  As an chipy.org admin - I'd like to maybe have a dashboard that shows how many job postings we had over a given period of time and how many of those were filled.  

I suggest we add the field and let the owner edit that field.  The postings list view would then have a filter such that approved, current, unfilled postings are displayed. 

I can take this ticket if you're good with the approach @JoeJasinski

Issue #358, opened 2021-02-11
Flat Pages should allow for specifiying custom images


raymondberg:
As a creator of a flat-page, I would like to be able to specify a custom/appropriate image for the header of my page. This could be a dropdown or manual entry for basic functionality. Images could be uploaded via the Admin, but more likely it would be good to have them committed to the source code like the other headers.

PR #594, opened 2025-02-21
Add Flat pack, and Job completed

HugoSeguin:
Completed Issue #573 To Mark a job post as filled + Issue #358 Flatpacks allowed for custom images. This is my first push so please look over it in depth. Hopefully it is up to your standards.


chipy_org/apps/job_board/views.py
```diff
@@ -169,7 +176,7 @@ class JobPostList(ListView):
     template_name = "job_board/job_post_list.html"
     ordering = ("-is_sponsor", "-approval_date")
     paginate_by = 8
-    queryset = JobPost.approved_and_active
+    queryset = JobPost.approved_and_active.filter(is_filled=False)
```
elmq0022, line 179: I'd think we'd update this method vs adding the filter here.

chipy_org/settings.py
```diff
@@ -61,6 +62,12 @@ def env_list(key, defaults=None, delimiter=","):
 
 MANAGERS = ADMINS
 
+
+BASE_DIR = Path(__file__).resolve().parent.parent
```
elmq0022, line 66: I probably have to try this out.  We do have media url and media root defined further down in the file.  It could be fine but these at least have to be in the same place in the file.

chipy_org/settings.py
```diff
@@ -239,6 +246,7 @@ def env_list(key, defaults=None, delimiter=","):
     "django_gravatar",
     "django_ical",
     "django_recaptcha",
+    "custom_flatpages",
```
elmq0022, line 249: Is this a third party app or just part of Django.

chipy_org/settings.py
```diff
@@ -334,8 +342,9 @@ def env_list(key, defaults=None, delimiter=","):
 NH3_ALLOWED_URL_SCHEMES = ["http", "https", "data"]
 NH3_STRIP_COMMENTS = True  # default, listed here for documentation
 
-RECAPTCHA_PUBLIC_KEY = env_var("NORECAPTCHA_SITE_KEY")
```
elmq0022, line 337: Please remove the recaptcha changes.  These won't work in prod.  There are test keys we can add to your local .env file that are specific to local dev and always pass. I will follow up on this separately.

elmq0022:
So I think the jub request part is pretty much on track.  I suggest putting that in a separate pull request so we can test and merge that functionality. We can handle updates to custom flatpages in a different PR.  

The image stuff for the flat pages will need a little more polish.  The first thing to do would be to move all the code into an app directory so it follows the structure of the rest of the project.

PR #595, opened 2025-03-06
Completed Issue #573 To Mark a job post as filled

HugoSeguin:


chipy_org/apps/job_board/models.py
```diff
@@ -10,11 +10,7 @@
 MAX_LENGTH = 100
 NUM_DAYS_T0_EXPIRE = 60
 
-STATUS_CHOICES = [
-    ("SU", "Submitted"),
-    ("AP", "Approved"),
-    ("RE", "Rejected"),
-]
+STATUS_CHOICES = [("SU", "Submitted"), ("AP", "Approved"), ("RE", "Rejected"), ("NA", "Filled")]
```
elmq0022, line 13: NIT: this isn't formatted correctly (the project uses the black formatting). Regardless the pre-commit hook will catch it.

chipy_org/apps/job_board/templates/job_board/job_post_detail.html
```diff
@@ -2,46 +2,42 @@
...
+    {% for job in jobs %}
+        <li>{{ job.title }} - 
+            {% if job.is_filled %}
+                Filled on {{ job.filled_date }}
+            {% else %}
+                Open
+            {% endif %}
+        </li>
+    {% endfor %}
+</ul>
```
elmq0022, line 17: I don't think this is what we want here.  This is the job post details page and should be related to a single post.  So I wouldn't expect a loop over jobs. Something like:

<p>
  <b>
     Available: 
        {% if job.is_filled %} 
           Filled on {{ job.filled_date }} 
       {% else %} 
           Yes 
        {% endif %}
   </b>
</p>

Though I'm still unsure why we would list filled jobs ... :thinking:

chipy_org/settings.py
```diff
@@ -31,7 +32,7 @@ def env_list(key, defaults=None, delimiter=","):
 
 PROJECT_ROOT = os.path.abspath(os.path.dirname(__file__))
 
-DEBUG = env_var("DEBUG", False)
+DEBUG = env_var("DEBUG", True)
```
elmq0022, line 35: This needs to be set back to False

chipy_org/settings.py
```diff
@@ -335,7 +342,8 @@ def env_list(key, defaults=None, delimiter=","):
 NH3_STRIP_COMMENTS = True  # default, listed here for documentation
 
 RECAPTCHA_PUBLIC_KEY = env_var("NORECAPTCHA_SITE_KEY")
-RECAPTCHA_PRIVATE_KEY = env_var("NORECAPTCHA_SECRET_KEY")
+RECAPTCHA_PRIVATE_KEY = env_var("NORECAPTCHA_PRIVATE_KEY")
```
elmq0022, line 345: Do not change these.  If you want to use your own local envs values that's fine but they need to be set via the docker env vars, not with a code change.

chipy_org/settings.py
```diff
@@ -335,7 +342,8 @@ def env_list(key, defaults=None, delimiter=","):
...
+SILENCED_SYSTEM_CHECKS = ["django_recaptcha.recaptcha_test_key_error"]
```
elmq0022, line 346: please remove this line.

chipy_org/apps/job_board/views.py
```diff
@@ -169,7 +181,7 @@ class JobPostList(ListView):
     template_name = "job_board/job_post_list.html"
     ordering = ("-is_sponsor", "-approval_date")
     paginate_by = 8
-    queryset = JobPost.approved_and_active
+    queryset = JobPost.approved_and_active.filter(is_filled=False)
```
elmq0022, line 184: I like this!

chipy_org/apps/job_board/templates/job_board/job_post_detail.html
elmq0022: Okay I get this now, but there's no limit or pagination here. What if we have a 1000+ filed jobs?   Unlikely I know, but we should put some limits like 10 or a pagination.  

Let's maybe just make this out of scope and remove it.

chipy_org/settings.py
```diff
@@ -2,6 +2,7 @@
 # Django settings for account project
```
elmq0022, line 2: Please revert the settings.py file.

chipy_org/templates/flatpages/default.html
```diff
@@ -1,7 +1,24 @@
 {% extends "shiny/slim.html" %}
-{% block page_header %}{{ flatpage.title | upper }}{% endblock %}
+
```
elmq0022, line 2: Please remove all the changes related to the flatpages and move them to a different PR as discussed.

chipy_org/urls.py
```diff
@@ -31,22 +32,22 @@
...
-            r"media/<path:path>",
+            "media/<path:path>",
             django.views.static.serve,
             {"document_root": settings.MEDIA_ROOT, "show_indexes": True},
         ),
     ]
     urlpatterns += staticfiles_urlpatterns()
+    urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
 handler404 = chipy_org.apps.main.views.custom_404
```
elmq0022, line 53: Please don't delete the docker.env.sample.  Other devs will find that useful when setting up a local env.

elmq0022:
I like the changes related to the job post model and the related migration. 

I didn't look into it, but how will django handle the old job posting in that don't have a filled status or filled date? 

Will users be able to edit their job filed status / date.  Is that available in the user facing edit form and not just the admin.  Most users will not have access to the admin.

Please remove the ability to see a list of filled jobs as the AC was optional.  I think it's fine to do, but we should have a separate template and view for that.  Let's skip for now so we can close this PR.

Please revert all other changes as noted.

eevelweezel:
Re: django and old posts...

Can we add logic in the migrations to set a status for the old records?

PR #596, opened 2025-03-06
https://github.com/chicagopython/chipy.org/issues/358

HugoSeguin:


elmq0022:
This seems to be the same as your other PR.  I'm going to to close this to avoid any confusion.

PR #601, opened 2025-04-18
Add `.PHONY` statements and PR template

michplunkett:
## Describe your changes
Added `.PHONY` statements above `Makefile` commands to disambiguate any commands from potential files in the repository. Added a PR template so that we have a standardized set of things to enter when submitting a PR.

## Checklist before requesting a review
- [x] `make format` has passed.
- [x] `make test` has passed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,67 +1,87 @@
+.PHONY: help
 help:
 	@echo "Type make, then hit tab to see make options"
 
+.PHONY: setup_env
 setup_env:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,67 +1,87 @@
+.PHONY: help
```
michplunkett, line 1: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

PR #603, opened 2025-08-21
Fix the contact form again

### LINKED ISSUE #577: Contact Form Appears to be Broken

eevelweezel:
Fixes #577 ... I hope.

Pretty sure the Crispy Form Helper was automagically overwriting the user-submitted email with the (empty) value from the social auth user profile.

Changes:
- remove crispy.FormHelper and convert to a regular Django form
- remove unused dep that angers cython
- tested locally
- unit tests are passing 
- screenshot from logs, but others are **_strongly_** encouraged to verify:

image

### DIFF
--- a/chipy_org/apps/contact/forms.py
+++ b/chipy_org/apps/contact/forms.py
@@ -15,7 +15,7 @@ class ContactForm(forms.Form):
         super().__init__(*args, **kwargs)
         self.fields["captcha"].label = False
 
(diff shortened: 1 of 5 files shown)


jimmerzzz52:
LGTM. Nice Tests!

PR #604, opened 2025-08-28
change field type and add email to body text

### LINKED ISSUE #577: Contact Form Appears to be Broken

eevelweezel:
Trying to fix #577, once again.

Change "email" field type to CharField and concat the value to the body text, just in case.

image

image


jimmerzzz52:
Actually LGTM, not like the other one.

PR #605, opened 2025-09-26
Switch from `runtime.txt` to `.python-version`

edmorley:
Since the `runtime.txt` file is deprecated:
https://devcenter.heroku.com/changelog-items/3141

In addition, I've:
- Updated `.python-version` to use the `3.13` version alias instead of an exact patch version, so that the latest patch version is automatically used. (3.13.1 is 9 months out of date and contains security vulnerabilities)
- Adjusted the GitHub Actions "Install Python" step to use the version from `.python-version`, saving the need to keep the versions in sync manually.

See:
- https://devcenter.heroku.com/changelog-items/3141
- https://devcenter.heroku.com/articles/python-runtimes
- https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-file-input


edmorley:
@eevelweezel Hi! I don't suppose you have a spare moment for a review/merge? :-)

eevelweezel:
Seems like a reasonable change.

